PR #7, opened 2023-11-02
feat: Add ConditionSet and status.Controller to emit logs, metrics, and events on status conditions.

ellistarn:
*Issue #, if available:*

*Description of changes:*

### ConditionSet

ConditionSet enables controllers to easily manage status conditions on their CRDs and encapsulates the complexity of condition initialization, transition timestamps, and condition rollup.

CRDs must implement a simple interface to be a "conditioned object":

```
func (t *TestObject) StatusConditions() status.ConditionSet {
	return status.NewReadyConditions(ConditionTypeFoo, ConditionTypeBar).For(t)
}

func (t *TestObject) GetConditions() []status.Condition {
	return t.Status.Conditions
}

func (t *TestObject) SetConditions(conditions []status.Condition) {
	t.Status.Conditions = conditions
}
```

A few concepts:

1. **Root Condition:** This is a top level condition (e.g. Ready, Succeeded) that is automatically set by the status of other conditions. It's true when dependent conditions are true, false when dependent conditions are false, and otherwise unknown.
2. **Normal Conditions:** These conditions are "well known" by the condition set, and will be present on all objects. Normal conditions rollup into the root condition. This is in contrast to "abnormal conditions", which may exist on objects, but do not impact the root condition. This is useful to provide information to users that may not indicate a problem with the operations of the object.

### Status Controller

Any object that implements `status.Object` can be used to create a `status.Controller`. This controller automatically monitors the conditions and condition transitions for status.Objects, and emits logs, events, and metrics. It's impossible to guarantee data consistency for these conditions, since Kubernetes (and metrics, for that matter) is an eventually consistent system. It's possible that transitions will be missed, or that the the state metrics will be delayed. 

### Notes

1. I intentionally didn't build support for observed generation, will follow up
2. I started with knative/pkg, but reoriented around the upstream metav1.Condition interfaces, removing things like `severity` as we need them. 


status/conditions.go
```diff
@@ -1,64 +1,68 @@
...
 
-type Conditions []v1.Condition
+type Conditions struct {
+	object     client.Object
+	conditions *[]Condition
+}
 
-func (c *Conditions) Set(conditionType string, conditionStatus v1.ConditionStatus, conditionMessage string) {
-	condition := c.Get(conditionType)
+func (c Conditions) Set(ctx context.Context, condition Condition) {
```
jonathan-innis, line 26: Do we want even more abstraction here similar to knative where we support things like `MarkTrue` and `MarkFalse` to make changing conditions simpler?

status/conditions.go
```diff
@@ -1,64 +1,68 @@
...
-type ConditionedObject interface {
-	v1.Object
-	runtime.Object
-	GetConditions() Conditions
+func NewConditions(object Object, conditions []Condition) Conditions {
+	return Conditions{object: object, conditions: lo.ToPtr(conditions)}
 }
 
-type Conditions []v1.Condition
+type Conditions struct {
```
jonathan-innis, line 21: So, to be clear, this is a "ConditionableObject". So, in order to initialize an object with this type, you have to wrap it. Is there a reason you decided to go this route as opposed to the interface route where you return a set of conditions from your object?

status/conditions.go
```diff
@@ -1,64 +1,68 @@
...
+	object     client.Object
+	conditions *[]Condition
```
jonathan-innis, line 23: Why is this a pointer to the conditions array?

status/conditions.go
```diff
@@ -1,64 +1,68 @@
...
-	metricNamespace            = "operator"
-	metricSubsystem            = "status_condition"
-)
-
-type ConditionedObject interface {
-	v1.Object
-	runtime.Object
-	GetConditions() Conditions
+func NewConditions(object Object, conditions []Condition) Conditions {
+	return Conditions{object: object, conditions: lo.ToPtr(conditions)}
```
jonathan-innis, line 18: Array should be pass-by-reference anyways so I'm unclear why you need a pointer here

status/conditions.go
```diff
@@ -1,64 +1,68 @@
...
-		})
+	existing := c.Get(condition.Type)
+	if existing == nil {
+		condition.LastTransitionTime = metav1.Now()
+		*c.conditions = append(*c.conditions, condition)
 		return
 	}
-
-	if condition.Status != conditionStatus {
+	if condition.Status != existing.Status {
```
jonathan-innis, line 35: I haven't looked at how knative did this but I'm assuming that if the reason value changes, that shouldn't dictate the change of the transition time?

status/conditions.go
```diff
@@ -1,64 +1,68 @@
...
+		condition.LastTransitionTime = metav1.Now()
+		// Record a metric
 		ConditionDuration.
```
jonathan-innis, line 38: What happens if the patch operation fails on updating this status condition? Will we fire the duration metric twice as well as event multiple times, which might skew the results

status/conditions.go
ellistarn: Added in rev2. This became clear when I started integrating with karpenter-core

status/conditions.go
ellistarn: Needed to be able to ref the object to add to events, logs, etc.

status/conditions.go
ellistarn: I need to be able to mutate the conditions on an object that this conditions struct manages. Otherwise it just mutates a copy of the slice.

status/conditions.go
ellistarn: Mutability

status/conditions.go
ellistarn: Per the spec, It doesn't count as a transition unless the status changes. Worth discussing if we agree on this.

status/conditions.go
ellistarn: Yes -- any thoughts on how to work around this?

object/object.go
```diff
@@ -17,37 +17,35 @@ import (
...
-			UID:                owner.GetUID(),
-			Controller:         lo.ToPtr(true),
-			BlockOwnerDeletion: lo.ToPtr(true),
-		}},
-	}
-	for _, meta := range mergeFrom {
-		lo.Must0(mergo.Merge(&object, &meta, mergo.WithOverride))
-	}
-	return object
+	owned.SetName(owner.GetName())
```
jonathan-innis, line 37: Why are we setting the name and namespace of the thing that we are owning from the owner? Maybe I'm missing what this is intending to do but I assumed that this would just update ownerReferences. 

Maybe a comment above the function could make it more clear what this library function is intending

status/condition_set.go
```diff
@@ -0,0 +1,295 @@
...
+				Type:     t,
+				Status:   lo.Ternary(cs.Root().IsTrue(), metav1.ConditionTrue, metav1.ConditionUnknown),
+				Severity: cs.severity(t),
+			})
+		}
+	}
+	return cs
+}
+
+// Root looks returns the root Condition
```
jonathan-innis, line 68: An example of what this is to someone who is unfamiliar? My assumption is that this is the `Ready` condition in almost all cases

status/controller.go
```diff
@@ -2,53 +2,158 @@ package status
...
+
+	objectLabels := prometheus.Labels{
+		MetricLabelGroup:     object.GVK(o).Group,
+		MetricLabelKind:      object.GVK(o).Kind,
+		MetricLabelNamespace: o.GetNamespace(),
+		MetricLabelName:      o.GetName(),
 	}
-	ConditionCount.DeletePartialMatch(labels)
 
+	ConditionCount.DeletePartialMatch(objectLabels)
```
jonathan-innis, line 62: We're going to take a gap in our metrics if we do it like this rather than updating if the objectLabels already exist and deleting the metric if the object labels don't exist.

It scares me a bit to do things this way since we don't persist data if our calls to the apiserver start to fail

status/controller.go
```diff
@@ -2,53 +2,158 @@ package status
...
+	return event.Event{
+		InvolvedObject: object,
+		Type:           string(lo.Ternary(condition.Severity == ConditionSeverityInfo && condition.Status == metav1.ConditionFalse, v1.EventTypeNormal, v1.EventTypeWarning)),
+		Reason:         string(condition.Type),
+		Message:        string(condition.Reason),
+	}
+}
+
+// Cardinality is limited to # objects * # conditions * # objectives
+var ConditionDuration = prometheus.NewSummaryVec(
```
jonathan-innis, line 118: How useful is this summary vector if it isn't summarizing any information? There's a bit of a conversation that happens around using summary vectors when you are trying to roll-up information. I'm curious how this is going to work if we are trying to roll-up a bunch of P99 information into a top-level P99? How are we going to do that? Are we going to average across the different elements from the different namespaces, weight them in some way? 

Ref: https://prometheus.io/docs/practices/histograms/#:~:text=interval%20of%20%CF%86.-,Two%20rules%20of%20thumb,-%3A

status/controller.go
jonathan-innis: There's an idea of how this might be handled that's present in a metric family implementation in `kubernetes-sigs/karpenter`: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/metrics/store.go#L25

status/controller.go
```diff
@@ -2,53 +2,158 @@ package status
...
+		forObject:          forObject,
+		observedConditions: map[reconcile.Request]ConditionSet{},
 	}
 }
 
 func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		For(c.forObject).
-		Named(fmt.Sprintf("metrics.%s", object.GVK(c.forObject).Kind)).
+		Named("status").
```
jonathan-innis, line 48: I'm assuming that we are calling all the controllers by the same name since controller-runtime can differentiate these controllers based on structured logging info?

status/conditions.go
jonathan-innis: Discussed offline and changed from tracking the metric within the state transition itself to watching the state transition from a different controller.

status/condition.go
```diff
@@ -0,0 +1,125 @@
...
+	ConditionSeverityError ConditionSeverity = ""
+	// ConditionSeverityInfo specifies that a failure of a condition type
+	// should be viewed as purely informational, and that things could still work.
+	ConditionSeverityInfo ConditionSeverity = "Info"
+)
+
+// Condition defines a readiness condition for a resource.
+// See: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
+// +k8s:deepcopy-gen=true
+type Condition struct {
```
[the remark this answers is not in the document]
ellistarn: I intentionally left this out, after I saw `status.ObservedGeneration` in knative pkg. I'm not sure which one is right, so I decided to punt on it for now.

object/object.go
ellistarn: Thanks for questioning this, I decided to just go w/ controller-runtime's helper in all cases, even if it isn't inlinable. Better than a weak opinion on top of it.

status/controller.go
ellistarn: Nice ref, will play with this.

status/controller.go
ellistarn: Maybe worth an offline discussion. I'm new to summary metrics, and may be missing something.

status/controller.go
ellistarn: Correct.

status/condition.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (c *Condition) GetStatus() metav1.ConditionStatus {
+	if c == nil {
+		return metav1.ConditionUnknown
+	}
+	return c.Status
+}
+
+// GetReason returns a nil save string of Reason
```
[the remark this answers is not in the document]
ellistarn: I think the original source meant "nil safe". 
Removed a bunch of these since they weren't used anywhere. We can reintroduce when we have a use case for the nil safe versions.

status/condition.go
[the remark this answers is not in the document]
ellistarn: Two differences: 

* Knative has the concept of `Severity`
* APIMachinery has the concept of observed generation
* APIMachinery requires a reason, where Knative has it as optional.

We're really only using severity for events right now (Info vs Warn). I could see us just doing info in all cases. 

All else equal, I think we probably want to just reuse the upstream concept wholesale. Maybe alias the type to add our helpers. WDYT?

test/object.go
```diff
@@ -15,7 +15,8 @@ import (
 )
 
 var (
-	DiscoveryLabel       = "operators.k8s.aws/test-id"
+	APIGroup             = "operators.k8s.aws"
```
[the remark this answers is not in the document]
ellistarn: I typically see singular if it's a brand, but plural if it's a concept. "operators for kubernetes at aws"

status/controller.go
```diff
@@ -4,51 +4,157 @@ import (
...
+		MetricLabelGroup,
+		MetricLabelKind,
+		MetricLabelNamespace,
+		MetricLabelName,
+		MetricLabelConditionType,
+		MetricLabelConditionStatus,
+	},
+)
+
+// Cardinality is limited to # objects * # conditions
```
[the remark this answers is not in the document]
ellistarn: Each condition can only be in one state at a time.

test/expectations.go
```diff
@@ -0,0 +1,73 @@
...
+	GinkgoHelper()
+	for _, o := range objects {
+		current := o.DeepCopyObject().(client.Object)
+		statuscopy := o.DeepCopyObject().(client.Object) // Snapshot the status, since create/update may override
+		// Create or Update
+		if err := c.Get(ctx, client.ObjectKeyFromObject(current), current); err != nil {
+			if errors.IsNotFound(err) {
+				Expect(c.Create(ctx, o)).To(Succeed())
+			} else {
+				Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
ellistarn: Done

status/condition_set.go
```diff
@@ -0,0 +1,244 @@
...
+			conditions = append(conditions, c)
+		} else {
+			// If we'd only update the LastTransitionTime, then return.
+			condition.LastTransitionTime = c.LastTransitionTime
+			if reflect.DeepEqual(condition, c) {
+				return
+			}
+		}
+	}
+	condition.LastTransitionTime = metav1.Now()
```
jonathan-innis, line 110: This means that the lastTransitionTime will update even if the only thing that changes is the message? Is that the correct behavior?

status/condition_set.go
```diff
@@ -0,0 +1,244 @@
...
+	if len(c.dependents) == 0 {
+		return nil
+	}
+
+	// Get dependent conditions
+	conditions := lo.Filter(c.object.GetConditions(), func(condition Condition, _ int) bool {
+		return lo.Contains(c.dependents, condition.Type)
+	})
+
+	// Sort set conditions by time.
```
jonathan-innis, line 204: How much do we care about what the dependent is that triggered the unhappiness for the ready condition? From what I can tell, we only care _that_ we have an unhappy dependent. Which one it is doesn't seem to matter as much

status/controller.go
```diff
@@ -6,49 +6,166 @@ import (
...
+	// lossy, specifically for when a condition transition rapidly. However,
+	// for the common case, we want to alert when a transition took a long
+	// time, and our likelyhood of observing this is much higher.
+	for _, condition := range currentConditions.List() {
+		observedCondition := observedConditions.Get(condition.Type)
+		if observedCondition == nil || observedCondition.GetStatus() == condition.GetStatus() {
+			continue
+		}
+		duration := condition.LastTransitionTime.Time.Sub(observedCondition.LastTransitionTime.Time).Seconds()
+		ConditionDuration.MustCurryWith(objectLabels).With(prometheus.Labels{
```
jonathan-innis, line 115: How big do our condition duration metrics get over the lifetime of the controller? We obviously shouldn't delete them after the object is gone, since we still need to scrape them to get accurate metrics over a given point of time; however, does this cardinality ever become a problem when it comes to memory usage since we are never deleting these counters?

status/condition.go
```diff
@@ -0,0 +1,58 @@
...
+package status
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Object interface {
+	client.Object
+	GetConditions() []Condition
```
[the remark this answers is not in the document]
ellistarn: I much prefer the explicit interface. It allows us to do things like manage pod conditions, as well, since we can adapt the conditiosn in these setters/getters.

status/condition_set.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// ConditionSet provides methods for evaluating Conditions.
+// +k8s:deepcopy-gen=false
+type ConditionSet struct {
+	ConditionTypes
+	object Object
+}
+
+// For creates a ConditionSet from an object using the original
```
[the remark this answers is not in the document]
ellistarn: Much rather have this be clear to callers, than within this impl.

test/expectations.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+const (
+	SlowTimeout = 100 * time.Second
+	SlowPolling = 10 * time.Second
+	FastTimeout = 1 * time.Second
+	FastPolling = 10 * time.Millisecond
+)
+
+func ExpectReconcile(ctx context.Context, controller reconcile.Reconciler, object client.Object) reconcile.Result {
```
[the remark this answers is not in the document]
ellistarn: Given that we almost always expect it to succeed, I'm thinking of making it `ExpectReconciled`, and we can skip having a helper for the failure case, since it's so rare. You can do `Expect(controller.Reconcile(nn)).ToNot(Succeed())`

object/object.go
```diff
@@ -2,61 +2,41 @@ package object
...
-		OwnerReferences: []metav1.OwnerReference{{
-			APIVersion:         gvk.GroupVersion().String(),
-			Kind:               gvk.Kind,
-			Name:               owner.GetName(),
-			UID:                owner.GetUID(),
-			Controller:         lo.ToPtr(true),
-			BlockOwnerDeletion: lo.ToPtr(true),
-		}},
+func (gvknn GroupVersionKindNamespacedName) String() string {
+	str := fmt.Sprintf("%s/%s", gvknn.Group, gvknn.Kind)
```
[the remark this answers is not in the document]
ellistarn: Not fussed for now.

status/doc.go
```diff
@@ -0,0 +1,3 @@
+// +k8s:deepcopy-gen=package,register
+// +kubebuilder:object:generate=false
```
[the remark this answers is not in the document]
ellistarn: Defaulting to false, and then I turn it on per struct. I didn't want to codegen the whole package.

status/controller.go
ellistarn: I suspect that many metrics have this problem. This problem exists whether or not we have the cardinality, since we are tracking individual data points for every transition. Every transition that occurs will grow the memory. I wonder if prometheus has a concept of whether or not a metric has been scraped such that the source (our process) can flush the memory. Maybe there's a TTL or something that we could use .

status/condition.go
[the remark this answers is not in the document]
jonathan-innis: Doesn't pod store the status conditions in the same place as other objects like `status.conditions`? Not sure that I get your point but I'm also fine with evolving this over time

status/condition_set.go
jonathan-innis: Seems fine to keep it this way for now

status/controller.go
```diff
@@ -6,49 +6,159 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	MetricLabelGroup           = "group"
+	MetricLabelKind            = "kind"
+	MetricLabelNamespace       = "namespace"
```
jonathan-innis, line 22: Drop label namespace and label name now that they are unused?

status/controller.go
```diff
@@ -6,49 +6,159 @@ import (
...
+	objectLabels := prometheus.Labels{
+		MetricLabelGroup: gvk.Group,
+		MetricLabelKind:  gvk.Kind,
 	}
-	ConditionCount.DeletePartialMatch(labels)
 
 	if err := c.kubeClient.Get(ctx, req.NamespacedName, o); err != nil {
-		return reconcile.Result{}, client.IgnoreNotFound(err)
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
```
jonathan-innis, line 66: We still need to delete the `ConditionCount`, right?

status/condition.go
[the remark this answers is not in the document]
ellistarn: It's not the same struct. It's similar, but predates the standard struct: https://pkg.go.dev/k8s.io/api/core/v1#PodCondition

status/condition_set.go
ellistarn: Ideally the unhappy dependent should roll up into the root condition. If there are multiple, we pick the first.

status/controller.go
ellistarn: Discussed offline. This is tricky, and it's not totally clear how histograms are aggregated over time. e.g., if we clear out a histogram metric, will that cause a spike in data? If the process restarts, is that equivalent to clearing it out? We decided to address this in a followon and remove NN from the duration metric for now.

status/controller.go
ellistarn: restored these for object counts. Keeping it in.

status/condition_set.go
jonathan-innis: Ah, I got it. That's the piece I missed. Then, I think that this makes sense.

jonathan-innis (approved):
LGTM 🚀

PR #11, opened 2024-02-22
feat: Accept any instead of client.Object

ellistarn:
*Issue #, if available:*

*Description of changes:*

Without this change, I'm forced to do crazy things like
```
	CRDs = []**apiextensionsv1.CustomResourceDefinition{
		object.Unmarshal[*apiextensionsv1.CustomResourceDefinition](bytes),
	}
```
because of
```
"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1".CustomResourceDefinition does not satisfy client.Object (method DeepCopyObject has pointer receiver)
```

### DIFF
--- a/object/object.go
+++ b/object/object.go
@@ -39,7 +39,7 @@ func GVK(o client.Object) schema.GroupVersionKind {
 	return lo.Must(apiutil.GVKForObject(o, scheme.Scheme))
 }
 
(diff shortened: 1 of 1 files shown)


jonathan-innis (approved):
LGTM 🚀

PR #12, opened 2024-02-22
fix: yaml parsing for crds

ellistarn:
*Issue #, if available:*

*Description of changes:*

Without this change:
```
  unable to install CRDs onto control plane: unable to create CRD instances: unable to get CRD "" to check if it exists: resource name may not be empty
```


jonathan-innis (approved):
LGTM 🚀

PR #17, opened 2024-03-20
Added context aliasing to simplify consumer imports

ellistarn:
*Issue #, if available:*

*Description of changes:*
Alias context so customers can just import a single context package.

Before

```
import (
        "context"
        operatorcontext "github.com/awslabs/operatorpkg/context"
)
func main() {
	ctx := operatorcontext.New()
	mgr := lo.Must(controllerruntime.NewManager(controllerruntime.GetConfigOrDie(), controllerruntime.Options{
		BaseContext:                   func() context.Context { return ctx },
	}))
}
```

After
```
import (
       "github.com/awslabs/operatorpkg/context"
)
func main() {
	ctx := context.New()
	mgr := lo.Must(controllerruntime.NewManager(controllerruntime.GetConfigOrDie(), controllerruntime.Options{
		BaseContext:                   func() context.Context { return ctx },
	}))
}
```

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"reflect"
 
-	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
 	"github.com/samber/lo"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


context/context.go
```diff
@@ -13,12 +12,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+type Context context.Context
```
jonathan-innis, line 15: It's not immediately obvious to me what this does. Can you show an example in the description of the PR?

ellistarn: Added an example

ellistarn: I'm essentially subclassing the context package with our own helpers.

jonathan-innis (approved):

PR #19, opened 2024-03-25
feat: Added a modified boolean to condition setters

ellistarn:
*Issue #, if available:*

*Description of changes:*
Adds a helper boolean to simplify caller code when modifying status conditions

This was inspired by `modified := controllerutil.SetFinalizer()`

Before
```
	if !obj.StatusConditions().Get(conditionType).IsTrue() {
		obj.StatusConditions().SetTrue(conditionType)
		if err := c.client.Status().Update(ctx, obj); err != nil {
			return reconcile.Result{}, err
		}
	}
```
After
```
	if obj.StatusConditions().SetTrue(conditionType) {
		if err := c.client.Status().Update(ctx, obj); err != nil {
			return reconcile.Result{}, err
		}
	}
```

### DIFF
--- a/status/condition_set.go
+++ b/status/condition_set.go
@@ -91,7 +91,7 @@ func (c ConditionSet) Get(t string) *Condition {
 
 // Set sets or updates the Condition on Conditions for Condition.Type.
 // If there is an update, Conditions are stored back sorted.
-func (c ConditionSet) Set(condition Condition) {
+func (c ConditionSet) Set(condition Condition) (modified bool) {
 	conditionType := condition.Type
 	var conditions []Condition
 	for _, c := range c.object.GetConditions() {
@@ -101,7 +101,7 @@ func (c ConditionSet) Set(condition Condition) {
 			// If we'd only update the LastTransitionTime, then return.
 			condition.LastTransitionTime = c.LastTransitionTime
 			if reflect.DeepEqual(condition, c) {
-				return
+				return false
 			}
 		}
 	}
@@ -113,6 +113,7 @@ func (c ConditionSet) Set(condition Condition) {
 
 	// Recompute the root condition after setting any other condition
 	c.recomputeRootCondition(conditionType)
+	return true
 }
 
 // RemoveCondition removes the non normal condition that matches the ConditionType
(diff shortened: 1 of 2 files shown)


status/condition_set.go
```diff
@@ -91,7 +91,7 @@ func (c ConditionSet) Get(t string) *Condition {
 
 // Set sets or updates the Condition on Conditions for Condition.Type.
 // If there is an update, Conditions are stored back sorted.
-func (c ConditionSet) Set(condition Condition) {
+func (c ConditionSet) Set(condition Condition) (modified bool) {
```
njtran, line 94: feels weird to name the return value if you're not going to return the named variable. Wouldn't the better way to inform what the function is supposed to return with a comment?

status/condition_set.go
```diff
@@ -146,14 +147,14 @@ func (c ConditionSet) Clear(t string) error {
 
 // SetTrue sets the status of t to true with the reason, and then marks the root condition to
 // true if all other dependents are also true.
-func (c ConditionSet) SetTrue(conditionType string) {
-	c.SetTrueWithReason(conditionType, conditionType, conditionType)
+func (c ConditionSet) SetTrue(conditionType string) (modified bool) {
```
njtran, line 150: same here
```suggestion
func (c ConditionSet) SetTrue(conditionType string) bool {
```

status/condition_set.go
```diff
@@ -163,16 +164,26 @@ func (c ConditionSet) SetTrueWithReason(conditionType string, reason, message st
...
+	return r.Set(Condition{
 		Type:    conditionType,
 		Status:  metav1.ConditionUnknown,
 		Reason:  "AwaitingReconciliation",
 		Message: "object is awaiting reconciliation",
 	})
 }
 
+// SetFalse sets the status of t and the root condition to False.
+func (r ConditionSet) SetFalse(conditionType string, reason, message string) (modified bool) {
```
njtran, line 178: nit: seems like we should match this function definition with the SetTrue() one since it takes in a reason
```suggestion
func (r ConditionSet) SetFalseWithReason(conditionType string, reason, message string) (modified bool) {
```

ellistarn: Not willing to make this breaking change right now.

status/condition_set.go
ellistarn: I prefer self documenting code than prose documentation

ellistarn: We also need to revisit the comments on this whole package at some point. Not sure how go docgen works with these.

PR #20, opened 2024-03-25
feat: Added conditions.IsTrue()

ellistarn:
*Issue #, if available:*

*Description of changes:*
Useful for controllers to wait upon each other.
```
if !obj.StatusConditions().IsTrue(DependsOnConditions...) {
	return reconcile.Result{}, nil
}
```


status/condition_set.go
```diff
@@ -89,6 +89,16 @@ func (c ConditionSet) Get(t string) *Condition {
 	return nil
 }
 
+// True returns true if all condition types are true.
+func (c ConditionSet) IsTrue(conditionTypes ...string) bool {
```
jonathan-innis, line 93: Curious if this semantically makes more sense to be "AreTrue()" or "AllTrue()" like you originally had it. If I were guessing at the function name, I'd probably go with AreTrue() because it's a plural.

jonathan-innis (approved):
One thought

status/condition_set.go
njtran: Since the type is a ConditionSet (a singular object/collection), then IsTrue() makes grammatical sense to me. 

I was also wondering, is it overfactored to just check the root conditions and make sure all root conditions are true? It really only has value if we think a dependent condition can be happy if it's false. For instance, if someone modeled their status conditions as a negative, like "Drifted", you'd want to know that the status condition is false to be considered happy.

ellistarn: I started with AllTrue, but didn't like the implications that you'd get true if you passed an empty variadic, when the condition set as a whole might be false. I then added inverted logic for the empty case, but found that even more confusing in my usage. 

Then I flipped to AreTrue, which felt awkward in the non plural case, which was most common. 

Then I chose IsTrue, as a slightly awkward alternative, but at least implies you're not looking at the condition set as a whole. "IsFoo" is also commonly used as a function that returns a Boolean, so it felt somewhat natural, even though the grammar is off for the plural variadic case.

ellistarn: I'm not grokking your question, Nick. There's only one root condition "Ready", and the rest are either normal or abnormal. Normal conditions are all true-positive, while abnormal can be anything.

PR #36, opened 2024-05-08
chore: Made status controller generic and split Apply expectations

ellistarn:
*Issue #, if available:*

*Description of changes:*

* Use a cleaner generic approach to the status controller
* Broke apart apply and status, since it broke for any async case (e.g. integ test) where the object was modified in flight. I couldn't just use PATCH since CRDs don't support strategic merge patches, so this approach was broken if we ever needed to write values like "" or 0. :(

### DIFF
--- a/object/object.go
+++ b/object/object.go
@@ -2,6 +2,7 @@ package object
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -37,6 +38,10 @@ func (gvknn GroupVersionKindNamespacedName) String() string {
 
 func GVK(o client.Object) schema.GroupVersionKind {
 	return lo.Must(apiutil.GVKForObject(o, scheme.Scheme))
+}
+
+func New[T any]() T {
+	return reflect.New(reflect.TypeOf(*new(T)).Elem()).Interface().(T)
 }
 
 func Unmarshal[T any](raw []byte) *T {
--- a/test/expectations/expectations.go
+++ b/test/expectations/expectations.go
@@ -66,7 +66,6 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
 	GinkgoHelper()
 	for _, o := range objects {
 		current := o.DeepCopyObject().(client.Object)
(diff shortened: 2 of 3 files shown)


test/expectations/expectations.go
```diff
@@ -78,18 +77,24 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
...
 		// Re-get the object to grab the updated spec and status
 		Expect(c.Get(ctx, client.ObjectKeyFromObject(o), o)).To(Succeed())
 	}
 }
 
+func ExpectStatusUpdated(ctx context.Context, c client.Client, objects ...client.Object) {
+	GinkgoHelper()
+	for _, o := range objects {
+		Expect(c.Status().Update(ctx, o.DeepCopyObject().(client.Object))).To(Succeed())
+		Expect(c.Get(ctx, client.ObjectKeyFromObject(o), o)).To(Succeed())
```
njtran, line 90: Wouldn't we want to get the object first then update it?

ellistarn: If we got the object first, we would override our desired object to be written, and the update would be a noop. 

We read afterwards to re-hydrate any mutating admission changes made to the object on the update, so we don't have a slightly-out-of-sync object. We also do this on create and delete, as the controller-runtime client only updates the resource version, and isn't aware of any of the server-side mutated fields.

object/object.go
```diff
@@ -39,6 +40,10 @@ func GVK(o client.Object) schema.GroupVersionKind {
 	return lo.Must(apiutil.GVKForObject(o, scheme.Scheme))
 }
 
+func New[T any]() T {
```
njtran, line 43: I like this better, abstract it away so I never have to think about this again !

test/expectations/expectations.go
njtran: Ah I wasn't aware of the re-hydration. That's smart. On getting the object first, @ellistarn and I synced offline. Looks like the trade-off here is using merge patches has the inability to nil out non-pointer strings, whereas update suffers from optimistic locking. If we can make sure we're doing a Get call before the Update, like we currently are, I think that's the better path.

njtran (dismissed):

njtran (approved):

PR #40, opened 2024-05-13
add expectation for status

njtran:
*Issue #, if available:*

*Description of changes:*
Adds a function to check to see if the expected conditions exist on the object

### DIFF
--- a/status/condition_set.go
+++ b/status/condition_set.go
@@ -158,7 +158,7 @@ func (c ConditionSet) Clear(t string) error {
 // SetTrue sets the status of t to true with the reason, and then marks the root condition to
 // true if all other dependents are also true.
 func (c ConditionSet) SetTrue(conditionType string) (modified bool) {
-	return c.SetTrueWithReason(conditionType, conditionType, conditionType)
+	return c.SetTrueWithReason(conditionType, conditionType, "")
 }
 
 // SetTrueWithReason sets the status of t to true with the reason, and then marks the root condition to
--- a/status/controller_test.go
+++ b/status/controller_test.go
@@ -56,6 +56,7 @@ var _ = Describe("Controller", func() {
 		time.Sleep(time.Second * 1)
 		testObject.StatusConditions().SetTrue(ConditionTypeFoo)
 		ExpectApplied(ctx, client, testObject)
+		ExpectObjectStatus(ctx, client, testObject, status.Condition{Type: ConditionTypeFoo, Status: metav1.ConditionTrue, Reason: ConditionTypeFoo})
 		ExpectReconciled(ctx, controller, testObject)
 
 		Expect(GetMetric("operator_status_condition_count", conditionLabels(status.ConditionReady, metav1.ConditionTrue))).To(BeNil())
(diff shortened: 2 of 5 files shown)


test/expectations/expectations.go
```diff
@@ -83,6 +84,18 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
 	}
 }
 
+func ExpectObjectStatus(ctx context.Context, c client.Client, obj status.Object, conditions ...status.Condition) {
```
ellistarn, line 87:
```suggestion
func ExpectStatusCondition(ctx context.Context, c client.Client, obj status.Object, conditions ...status.Condition) {
```

njtran: Did ExpectStatus() since we can have zero, one, or multiple status conditions.

test/expectations/expectations.go
```diff
@@ -83,6 +84,18 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
 	}
 }
 
+func ExpectStatus(ctx context.Context, c client.Client, obj status.Object, conditions ...status.Condition) {
```
ellistarn, line 87: I suggested ExpectStatusCondition, since there may be other aspects of status we need to verify that may not map well to this function, which specifically checks statusConditions

ellistarn (dismissed):

status/controller_test.go
```diff
@@ -106,7 +108,7 @@ var _ = Describe("Controller", func() {
 		Expect(GetMetric("operator_status_condition_transition_seconds", conditionLabels(ConditionTypeBar, metav1.ConditionUnknown)).GetHistogram().GetSampleCount()).To(BeNumerically(">", 0))
 
 		Expect(recorder.Events).To(Receive(Equal("Normal Bar Status condition transitioned, Type: Bar, Status: Unknown -> True, Reason: reason, Message: message")))
-		Expect(recorder.Events).To(Receive(Equal("Normal Ready Status condition transitioned, Type: Ready, Status: Unknown -> True, Reason: Ready, Message: Ready")))
+		Expect(recorder.Events).To(Receive(Equal("Normal Ready Status condition transitioned, Type: Ready, Status: Unknown -> True, Reason: Ready, Message: ")))
```
ellistarn, line 111: Is this event string right?

status/condition_set.go
```diff
@@ -158,7 +158,7 @@ func (c ConditionSet) Clear(t string) error {
 // SetTrue sets the status of t to true with the reason, and then marks the root condition to
 // true if all other dependents are also true.
 func (c ConditionSet) SetTrue(conditionType string) (modified bool) {
-	return c.SetTrueWithReason(conditionType, conditionType, conditionType)
+	return c.SetTrueWithReason(conditionType, conditionType, "")
```
jonathan-innis, line 161: I don't think this actually works. When I validate this in my environment, this fails due to a CRD validation error for not specifying a required field.

jonathan-innis (changes requested):

status/condition_set.go
jonathan-innis: Nvm, I'm wrong here. Tested again and this does in fact work.

status/controller_test.go
jonathan-innis: +1, seems like we should omit the message key if the value is empty

test/expectations/expectations.go
```diff
@@ -83,6 +84,18 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
 	}
 }
 
+func ExpectStatusConditions(ctx context.Context, c client.Client, obj status.Object, conditions ...status.Condition) {
```
jonathan-innis, line 87: How useful is this? Do we need to check exact values or can we just expect testers that are creating assertions to just validate the status of the condition?

jonathan-innis: On top of this, someone could just write `Expect(conditions, HasElements(...))` rather than using this function

ellistarn, line 87: Now that I look at this, why does this even take a context and client? Did you mean to call ExpectGet?

njtran: You don't have to care about observedGeneration or LastTransitionTime, so doing hasElements doesn't work since it's not strict equality

ellistarn (approved):

PR #45, opened 2024-05-17
feat: ExpectCleanedUp

ellistarn:
*Issue #, if available:*

*Description of changes:*

Will address finalizers in a followon

### DIFF
--- a/test/expectations/expectations.go
+++ b/test/expectations/expectations.go
@@ -3,12 +3,15 @@ package test
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/awslabs/operatorpkg/object"
 	"github.com/awslabs/operatorpkg/status"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/expectations/expectations.go
```diff
@@ -123,3 +126,22 @@ func ExpectDeleted(ctx context.Context, c client.Client, objects ...client.Objec
 		Expect(c.Get(ctx, client.ObjectKeyFromObject(o), o)).To(Or(Succeed(), MatchError(ContainSubstring("not found"))))
 	}
 }
+
+func ExpectCleanedUp(ctx context.Context, c client.Client, objects ...client.Object) {
```
njtran, line 130: Do we want to add an ExpectNotFound at the end to make sure it's all cleaned up before the function ends?

ellistarn: I can't call ExpectNotFound, because I'm acting on an object list -- there's no way to assert "expect none exist" without passing an object list, which I was struggling with.

njtran (approved):

PR #46, opened 2024-05-20
Add method to get status condition from unstructured object

jigisha620:
*Issue #, if available:*

*Description of changes:*
This PR adds a method to retrieve status conditions from an unstructured object. Making this change in response to - https://github.com/kubernetes-sigs/karpenter/pull/1248#discussion_r1605242106


status/condition_set.go
```diff
@@ -237,3 +238,27 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
 	})
 	return conditions
 }
+
+// GetConditionForUnstructured gets the condition for the given conditionPath. It returns an error if the condition is not found.
+func GetConditionForUnstructured(conditionPath string, nodeClass *unstructured.Unstructured) (*Condition, error) {
```
ellistarn, line 243: It looks like the majority of this code is converting between a runtime.Object and a status.Object. WDYT about instead making a helper like:

```
package status

func ObjectFrom(runtime.Object) {
   ...
}
```

And then we can reuse the standard condition handling methods?

jigisha620: Agreed. Added that change.

status/condition_set.go
```diff
@@ -237,3 +239,58 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
 	})
 	return conditions
 }
+
+// Unstructured is an embedded struct type that embeds the upstream type
+type Unstructured struct {
```
jonathan-innis, line 244: I liked @ellistarn's solution of having a method with the signature `FromUnstructured(obj unsturctured.Unstructured) status.Object` (or similar) to convert the existing type into a status object. That way, you don't have to expose an external type, since you will just be returning an interface that is already well-known. You will have to create an internal type inside of the function to get what you want, but this is really just syntactic sugar so that you don't have to export another type from this package.

jonathan-innis, line 244:
```suggestion
type objectImpl struct {
  conditions []metav1.Condition
  }
```

Consider defining your structured type as an implementation of the `status.Obejct` interface. This object stores `[]metav1.Condition`s inside of it. It also doesn't need to be exported, since you will just use it when you are trying to convert an object from `Unstructured` into something that implements `status.Object`

status/condition_set_test.go
```diff
@@ -105,4 +107,49 @@ var _ = Describe("Conditions", func() {
...
+		}
+		status.FromUnstructured(testObject).SetConditions(conditions)
+		c, found, err := unstructured.NestedSlice(testObject.Object, "status", "conditions")
+		Expect(err).To(BeNil())
+		Expect(found).To(BeTrue())
+		Expect(len(c)).To(BeEquivalentTo(2))
+	})
 })
+
+func createStatusCondition(status string) map[string]interface{} {
```
jonathan-innis, line 141:
```suggestion
func createUnstructuredStatusConditions(status string) map[string]interface{} {
```

jonathan-innis:
```suggestion
func createStatusCondition(status string) unstructured.Unstructured {
```

Should this return `unstructured.Unstructured` to more closely model the code behavior? I think it's just a couple more lines that you are adding in each test case anyways

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
 	})
 	return conditions
 }
+
+// objectImpl is an embedded struct type that embeds the upstream type
+type objectImpl struct {
```
jonathan-innis:
```suggestion
type unstructuredAdapter struct {
```

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+	unstructured.Unstructured
```
jonathan-innis, line 245:
```suggestion
	*unstructured.Unstructured
```

Should we store this as a pointer since we are receiving a pointer from the `FromUnstructured` call?

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+}
+
+func FromUnstructured(nodeClass *unstructured.Unstructured) Object {
```
jonathan-innis, line 248:
```suggestion
func FromUnstructured(unstructured *unstructured.Unstructured) Object {
```

nit

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
 	})
 	return conditions
 }
+
+// objectImpl is an embedded struct type that embeds the upstream type
```
jonathan-innis: Can we document that this assumes that you have your status conditions sitting on the `status.conditions` path?  I think this is a common enough paradigm that we can pretty much assume that conditions will _always_ sit on this path, but the adapter that we are using here (as well as the `FromUnstructured` function) should document that this is the case

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+	unstructured.Unstructured
+}
+
+func FromUnstructured(nodeClass *unstructured.Unstructured) Object {
+	return &objectImpl{Unstructured: *nodeClass}
+}
+
+func (o *objectImpl) GetConditions() []Condition {
+	var conditions []Condition
+	c, found, err := unstructured.NestedSlice(o.Object, "status", "conditions")
```
jonathan-innis: I wonder if we should validate this as part of the constructor. That way you don't have a chance to fail silently later. You get pre-validated that the path exists and that there are conditions present

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+	if err != nil || !found {
+		return conditions
+	}
+	for _, condition := range c {
+		cond := condition.(map[string]interface{})
+		newCond := &Condition{
+			Type:    cond["type"].(string),
+			Status:  metav1.ConditionFalse,
```
jonathan-innis, line 262: Can you do an inline conversion to a boolean and then convert this into a `metav1.Status`?

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+}
+
+func (o *objectImpl) GetConditions() []Condition {
+	var conditions []Condition
+	c, found, err := unstructured.NestedSlice(o.Object, "status", "conditions")
+	if err != nil || !found {
+		return conditions
+	}
+	for _, condition := range c {
+		cond := condition.(map[string]interface{})
```
jonathan-innis, line 259: Is there a way to perform a one-liner or some typed conversion into a `metav1.Condition` rather than having to grab the key from each field?

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+		}
+		conditions = append(conditions, *newCond)
+	}
+	return conditions
+}
+
+func (o *objectImpl) SetConditions(conditions []Condition) {
+	c, found, err := unstructured.NestedSlice(o.Object, "status", "conditions")
+	if err == nil && found {
+		for _, cond := range conditions {
```
jonathan-innis, line 277: `lo.Map()`?

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+			newCond.Status = metav1.ConditionTrue
+		}
+		conditions = append(conditions, *newCond)
+	}
+	return conditions
+}
+
+func (o *objectImpl) SetConditions(conditions []Condition) {
+	c, found, err := unstructured.NestedSlice(o.Object, "status", "conditions")
+	if err == nil && found {
```
jonathan-innis, line 276: Can we guarantee this through our constructor like we do in `GetConditions()`?

status/condition_set.go
```diff
@@ -237,3 +239,60 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+			}
+			c = append(c, newCondition)
+		}
+		unstructured.SetNestedSlice(u.Object, c, "status", "conditions")
+	}
+}
+
+func (u *unstructuredAdapter) StatusConditions() ConditionSet {
+	var conditionType []string
+	for _, c := range u.GetConditions() {
```
jonathan-innis, line 294: `lo.Map()`?

status/condition_set.go
```diff
@@ -237,3 +240,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
 	})
 	return conditions
 }
+
+// unstructuredAdapter is an embedded struct type that embeds the upstream type. unstructuredAdapter
```
jonathan-innis, line 244:
```suggestion
// unstructuredAdapter is an adapter for the status.Object interface. unstructuredAdapter
```

status/condition_set.go
```diff
@@ -237,3 +240,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+type unstructuredAdapter struct {
+	*unstructured.Unstructured
+}
+
+// FromUnstructured makes the assumption that the status conditions are found on status.conditions
+// path. If they are not found then we return nil.
+func FromUnstructured(u *unstructured.Unstructured) Object {
+	_, found, err := unstructured.NestedSlice(u.Object, "status", "conditions")
+	if err != nil || !found {
+		return nil
```
jonathan-innis, line 255: I think we should return an error here. So `FromUnstructured` has the signature (Object, error) for return arguments

status/condition_set.go
```diff
@@ -237,3 +240,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+	return &unstructuredAdapter{Unstructured: u}
+}
+
+func (u *unstructuredAdapter) GetConditions() []Condition {
+	var conditions []Condition
+	c, _, _ := unstructured.NestedSlice(u.Object, "status", "conditions")
+	for _, condition := range c {
+		var newCondition Condition
+		cond := condition.(map[string]interface{})
+		jsonStr, err := json.Marshal(cond)
```
jonathan-innis, line 266: Should we perform validation on these things ahead of time in the constructor path so that it's impossible for us to fail on these methods inside of the interface here. Silent failure like this is dubious because you drop data without telling the user why you have dropped the data.

We can't return an error here because we wouldn't be implementing the interface if we did, so we should return an error when constructing by performing pre-validation on the data that we are manipulating

status/condition_set.go
```diff
@@ -237,3 +240,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+		conditions = append(conditions, newCondition)
+	}
+	return conditions
+}
+
+func (u *unstructuredAdapter) SetConditions(conditions []Condition) {
+	uc, _, _ := unstructured.NestedSlice(u.Object, "status", "conditions")
+	lo.Map(conditions, func(condition Condition, _ int) []interface{} {
+		uc = append(uc, map[string]interface{}{
+			"type":               condition.Type,
```
jonathan-innis, line 282: Use the `metav1.Condition` and convert it to `map[string]interface{}?

status/condition_set.go
```diff
@@ -237,3 +240,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+			"lastTransitionTime": condition.LastTransitionTime.Time.String(),
+		})
+		return uc
+	})
+	unstructured.SetNestedSlice(u.Object, uc, "status", "conditions")
+}
+
+func (u *unstructuredAdapter) StatusConditions() ConditionSet {
+	var conditionType []string
+	lo.Map(u.GetConditions(), func(condition Condition, _ int) []string {
```
jonathan-innis, line 296:
```suggestion
	conditionTypes := lo.Map(u.GetConditions(), func(condition Condition, _ int) []string {
```

lo.Map outputs a new array so that you don't have to pre-declare and append within the function itself. You just return what you want the elements to look like in the array and `lo` constructs it all for you

status/condition_set.go
jigisha620: Agreed :thumbsup:

status/condition_set.go
```diff
@@ -237,3 +239,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+type unstructuredAdapter struct {
+	*unstructured.Unstructured
+}
+
+// FromUnstructured makes the assumption that the status conditions are found on status.conditions
+// path. If they are not found then we return nil.
+func FromUnstructured(u *unstructured.Unstructured) (Object, error) {
+	c, found, err := unstructured.NestedSlice(u.Object, "status", "conditions")
+	if err != nil || !found {
+		return nil, fmt.Errorf("unable to determine status conditions: %w", err)
```
ellistarn, line 254: Thoughts on panicking, since we control the types that we call this with? This will simplify callers significantly.

ellistarn (approved):

status/condition_set.go
jonathan-innis: I'm wondering if since this is a library if we should just be comfortable with throwing an error here and then letting the caller control whether they want to panic or not. I'm realistically fine with either; just thinking that errors give you a bit more flexibility

status/condition_set.go
```diff
@@ -237,3 +239,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+	}
+	for _, condition := range c {
```
jonathan-innis, line 256: nit: Consider commenting above this with something like, "validating status conditions are `metav1.Condition` types so it's a bit quicker to grok

status/condition_set.go
```diff
@@ -237,3 +239,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+			return nil, err
+		}
+	}
+	return &unstructuredAdapter{Unstructured: u}, nil
+}
+
+func (u *unstructuredAdapter) GetConditions() []Condition {
+	var conditions []Condition
+	c, _, _ := unstructured.NestedSlice(u.Object, "status", "conditions")
+	for _, condition := range c {
```
jonathan-innis, line 273: `lo.Map`?

status/condition_set.go
```diff
@@ -237,3 +239,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+func (u *unstructuredAdapter) SetConditions(conditions []Condition) {
+	uc, _, _ := unstructured.NestedSlice(u.Object, "status", "conditions")
+	lo.Map(conditions, func(condition Condition, _ int) []interface{} {
+		var b map[string]interface{}
+		j, _ := json.Marshal(&condition)
+		json.Unmarshal(j, &b)
+		uc = append(uc, b)
+		return uc
+	})
+	unstructured.SetNestedSlice(u.Object, uc, "status", "conditions")
```
jonathan-innis, line 292:
```suggestion
	unstructured.SetNestedSlice(u.Object, lo.Map(conditions, func(condition Condition, _ int) []interface{} {
		var b map[string]interface{}
		j, _ := json.Marshal(&condition)
		json.Unmarshal(j, &b)
		return b
	}), "status", "conditions")
```

nit: consider the following

status/condition_set.go
```diff
@@ -237,3 +239,62 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
...
+}
+
+func (u *unstructuredAdapter) StatusConditions() ConditionSet {
+	conditionType := lo.Map(u.GetConditions(), func(condition Condition, _ int) string {
```
jonathan-innis, line 296:
```suggestion
	conditionTypes := lo.Map(u.GetConditions(), func(condition Condition, _ int) string {
```

nit: considering making it plural because it is a plural thing

jonathan-innis (approved):
LGTM 🚀 Nice work!

status/condition_set.go
jigisha620: > Thoughts on panicking, since we control the types that we call this with? This will simplify callers significantly.

I have seen that the caller gets a bit complicated if we don't panic since there's some error handling that has to be done. 
But I also agree with @jonathan-innis that caller can decide what needs to be done in case of an error.

ellistarn (dismissed):

status/condition_set_test.go
```diff
@@ -105,4 +107,63 @@ var _ = Describe("Conditions", func() {
...
+			"invalid": map[string]interface{}{
+				"invalid": "invalid",
+			},
+		}}
+		obj, err := status.FromUnstructured(testObject)
+		Expect(obj).To(BeNil())
+		Expect(err).To(HaveOccurred())
+	})
+	It("should validate status condition on unstructured object status is false", func() {
+		unstructuredAdapter, err := status.FromUnstructured(createUnstructuredStatusConditions("False"))
```
jonathan-innis, line 121:
```suggestion
		conditionObj, err := status.FromUnstructured(createUnstructuredStatusConditions("False"))
```

nit

status/condition_set_test.go
```diff
@@ -105,4 +107,63 @@ var _ = Describe("Conditions", func() {
...
+		Expect(err).To(BeNil())
+		Expect(unstructuredAdapter).To(Not(BeNil()))
+		Expect(unstructuredAdapter.StatusConditions().IsTrue(status.ConditionReady)).To(BeFalse())
+	})
+	It("should validate status condition on unstructured object status is true", func() {
+		unstructuredAdapter, err := status.FromUnstructured(createUnstructuredStatusConditions("True"))
```
jonathan-innis, line 127:
```suggestion
		conditionObj, err := status.FromUnstructured(createUnstructuredStatusConditions("True"))
```

nit

status/condition_set_test.go
```diff
@@ -105,4 +107,63 @@ var _ = Describe("Conditions", func() {
...
+		Expect(err).To(BeNil())
+		Expect(unstructuredAdapter).To(Not(BeNil()))
```
jonathan-innis, line 129:
```suggestion
		Expect(unstructuredAdapter).ToNot(BeNil())
```

nit

status/condition_set_test.go
```diff
@@ -105,4 +107,63 @@ var _ = Describe("Conditions", func() {
...
+			},
+		}}
+		obj, err := status.FromUnstructured(testObject)
+		Expect(obj).To(BeNil())
+		Expect(err).To(HaveOccurred())
+	})
+	It("should validate status condition on unstructured object status is false", func() {
+		unstructuredAdapter, err := status.FromUnstructured(createUnstructuredStatusConditions("False"))
+		Expect(err).To(BeNil())
+		Expect(unstructuredAdapter).To(Not(BeNil()))
```
jonathan-innis, line 123:
```suggestion
		Expect(unstructuredAdapter).ToNot(BeNil())
```

nit

status/condition_set_test.go
```diff
@@ -105,4 +107,63 @@ var _ = Describe("Conditions", func() {
...
+		}}
+		conditions := []status.Condition{
+			{
+				Type:    status.ConditionSucceeded,
+				Status:  metav1.ConditionFalse,
+				Reason:  "reason",
+				Message: "message",
+			},
+		}
+		unstructuredAdapter, err := status.FromUnstructured(testObject)
```
jonathan-innis, line 146:
```suggestion
		conditionObj, err := status.FromUnstructured(testObject)
```

status/condition_set_test.go
```diff
@@ -105,4 +107,63 @@ var _ = Describe("Conditions", func() {
...
+		Expect(err).To(BeNil())
```
jonathan-innis, line 147:
```suggestion
		Expect(err).ToNot(HaveOccurred())
```

nit: You seem to use this style elsewhere, so consider keeping it for consistency

jonathan-innis (dismissed):
Bunch of nits, lgtm 🚀

jonathan-innis (approved):
LGTM 🚀

jonathan-innis:
Closing this out since we went a different direction. We'll re-open this if we think that this is needed later down the line.

PR #51, opened 2024-06-05
chore: Add controller interface and Requeue Immediately concepts

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

Add the `controller.Controller` interface and `singleton.RequeueImmediately` concepts

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -1,0 +1,12 @@
+package controller
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// Controller is a reconciler that allows registration with a controller-runtime Manager
+type Controller interface {
+	Register(context.Context, manager.Manager) error
+}
--- a/singleton/controller.go
+++ b/singleton/controller.go
@@ -2,12 +2,19 @@ package singleton
 
 import (
 	"context"
+	"time"
 
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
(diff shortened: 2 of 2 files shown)


singleton/controller.go
```diff
@@ -10,6 +11,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	// RequeueImmediately is a constant that allows for immediate RequeueAfter when you want to run your
+	// singleton controller as hot as possible in a fast requeuing loop
+	RequeueImmediately = 1 * time.Nanosecond
```
ellistarn, line 17: singleton.RequeueImmediately or controller.RequeueImediately?

This is to avoid backoff retry with Requeue: true, right?

jonathan-innis: Figured singleton might make more sense since you probably don't want to run a standard controller that hot

jonathan-innis: And putting it in controller might suggest that there are cases to do so

ellistarn: Why wouldn't you requeue: True for a regular controller?

ellistarn: I think there are plenty of cases to requeue a regular controller

controller/controller.go
```diff
@@ -0,0 +1,12 @@
+package controller
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// Controller is a reconciler that allows registration with a controller-runtime Manager
+type Controller interface {
```
ellistarn, line 10: Is this ~= Runnable from controller-runtime?

ellistarn: Or Registrable

jonathan-innis: Runnable implements Start() so it's slightly different. Registerable feels more awkward and perhaps less direct for what this intended to be used for. We could enforce that the controller is also a Reconciler as well if we wanted to make it a more tightly defined concept, but I'm not fully convinced that's necessary here. 

I personally still like calling this Controller.

ellistarn: I guess you register controllers with the controller managers. SGTM

ellistarn (approved):

PR #62, opened 2024-07-30
chore: Add `operator_status_condition_current_status_seconds` and `operator_status_condition_transitions_total`

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

Add the `operator_status_condition_current_status_seconds` and `operator_status_condition_transitions_total`.

#### `operator_status_condition_current_status_seconds`

This metric lets us know an approximate amount of time that a condition has been in a given state. This helps us identify cases where a status condition has not transitioned out of a state within a reasonable amount of time or if an abnormal status condition hasn't been acted-on quickly enough

#### `operator_status_condition_transitions_total`

This metric lets us know the total number of transitions that occur for a condition. We can aggregate this metric across an object kind to understand the number of objects that have transitioned.


status/controller.go
```diff
@@ -78,21 +88,41 @@ func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (r
...
 		}).Set(1)
+		ConditionCurrentStatusSeconds.With(prometheus.Labels{
+			MetricLabelGroup:           gvk.Group,
+			MetricLabelKind:            gvk.Kind,
+			MetricLabelNamespace:       req.Namespace,
+			MetricLabelName:            req.Name,
+			MetricLabelConditionType:   condition.Type,
+			MetricLabelConditionStatus: string(condition.Status),
+			MetricLabelConditionReason: condition.Reason,
+		}).Set(time.Since(condition.LastTransitionTime.Time).Seconds())
```
njtran, line 105: Any reason we don't want milliseconds here? Just too granular?

status/controller.go
```diff
@@ -114,25 +144,36 @@ func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (r
 	// time, and our likelyhood of observing this is much higher.
 	for _, condition := range currentConditions.List() {
 		observedCondition := observedConditions.Get(condition.Type)
-		if observedCondition == nil || observedCondition.GetStatus() == condition.GetStatus() {
+		if observedCondition.GetStatus() == condition.GetStatus() {
```
njtran, line 147: Do we want to treat it as a transition if it's the same status but a different reason?

njtran (approved):
LGTM just two comments

status/controller.go
jonathan-innis: No, I don't think that this should be considered a transition. We should consider a transition to be a status True/False change

status/controller.go
jonathan-innis: Yeah, I think milliseconds is too granular. Given that monitoring will most likely be on something that is stuck for an extended period of time (many, many seconds), I think seconds makes the most sense.

njtran (approved):

PR #67, opened 2024-08-09
add support for observedGeneration

wangamel:
*Issue #, if available:*

*Description of changes:*
This is to add the observedGeneration for each condition

### DIFF
--- a/status/condition_set.go
+++ b/status/condition_set.go
@@ -103,31 +103,21 @@ func (c ConditionSet) IsTrue(conditionTypes ...string) bool {
 // If there is an update, Conditions are stored back sorted.
 func (c ConditionSet) Set(condition Condition) (modified bool) {
 	conditionType := condition.Type
-	skipResetLastTransitionTime := false
 	var conditions []Condition
 	for _, cond := range c.object.GetConditions() {
 		if cond.Type != conditionType {
 			conditions = append(conditions, cond)
 		} else {
-			condition.LastTransitionTime = cond.LastTransitionTime
-			condition.ObservedGeneration = cond.ObservedGeneration
+			condition.LastTransitionTime = metav1.Now()
+			condition.ObservedGeneration = c.object.GetGeneration()
+			if condition.Status == cond.Status && cond.LastTransitionTime.UnixNano() > 0 {
+				condition.LastTransitionTime = cond.LastTransitionTime
+			}
 			if reflect.DeepEqual(condition, cond) {
-				// If we'd only update the LastTransitionTime, then return.
-				if cond.ObservedGeneration == c.object.GetGeneration() {
-					return false
-				} else {
-					// If we'd only update the ObservedGeneration, then keep the LastTransitionTime.
-					skipResetLastTransitionTime = true
-					condition.LastTransitionTime = cond.LastTransitionTime
-					condition.ObservedGeneration = c.object.GetGeneration()
-				}
+				return false
 			}
 		}
 	}
-	if !skipResetLastTransitionTime {
-		condition.LastTransitionTime = metav1.Now()
-	}
-	condition.ObservedGeneration = c.object.GetGeneration()
 	conditions = append(conditions, condition)
 	// Sorted for convenience of the consumer, i.e. kubectl.
 	sort.Slice(conditions, func(i, j int) bool { return conditions[i].Type < conditions[j].Type })
@@ -221,9 +211,12 @@ func (c ConditionSet) recomputeRootCondition(conditionType string) {
 	} else {
 		c.Set(Condition{
 			Type: c.root,
-			// The root condition is no longer unknown as soon as any are false
+			// The root condition is no longer unknown as soon as any dependency condition false with latest observedGeneration
 			Status: lo.Ternary(
-				lo.ContainsBy(conditions, func(condition Condition) bool { return condition.IsFalse() }),
+				lo.ContainsBy(conditions, func(condition Condition) bool {
+					return condition.IsFalse() &&
+						condition.ObservedGeneration == c.object.GetGeneration()
+				}),
 				metav1.ConditionFalse,
 				metav1.ConditionUnknown,
 			),
@@ -245,7 +238,7 @@ func (c ConditionSet) findUnhealthyDependents() []Condition {
 		return lo.Contains(c.dependents, condition.Type)
 	})
 	conditions = lo.Filter(conditions, func(condition Condition, _ int) bool {
-		return condition.IsFalse() || condition.IsUnknown()
+		return condition.IsFalse() || condition.IsUnknown() || condition.ObservedGeneration != c.object.GetGeneration()
 	})
 
 	// Sort set conditions by time.
--- a/status/condition_set.go
+++ b/status/condition_set.go
@@ -104,20 +104,27 @@ func (c ConditionSet) IsTrue(conditionTypes ...string) bool {
 func (c ConditionSet) Set(condition Condition) (modified bool) {
 	conditionType := condition.Type
 	var conditions []Condition
-	for _, c := range c.object.GetConditions() {
-		if c.Type != conditionType {
-			conditions = append(conditions, c)
+	for _, cond := range c.object.GetConditions() {
+		if cond.Type != conditionType {
+			conditions = append(conditions, cond)
 		} else {
-			// If we'd only update the LastTransitionTime, then return.
-			condition.LastTransitionTime = c.LastTransitionTime
-			condition.ObservedGeneration = c.ObservedGeneration
-			if reflect.DeepEqual(condition, c) {
-				return false
+			condition.LastTransitionTime = cond.LastTransitionTime
+			condition.ObservedGeneration = cond.ObservedGeneration
+			if reflect.DeepEqual(condition, cond) {
+				// If we'd only update the LastTransitionTime, then return.
+				if cond.ObservedGeneration == c.object.GetGeneration() {
+					return false
+				} else {
+					// If we'd only update the ObservedGeneration, then keep the LastTransitionTime.
+					condition.LastTransitionTime = cond.LastTransitionTime
+					condition.ObservedGeneration = c.object.GetGeneration()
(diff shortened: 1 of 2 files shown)


status/condition_set.go
```diff
@@ -110,12 +110,14 @@ func (c ConditionSet) Set(condition Condition) (modified bool) {
 		} else {
 			// If we'd only update the LastTransitionTime, then return.
 			condition.LastTransitionTime = c.LastTransitionTime
+			condition.ObservedGeneration = c.ObservedGeneration
```
jonathan-innis, line 113: So if the object generation changes here, I'm not sure that we can just bail, right? We have to always update to what the current object generation is here. So, if the object generation changes and we still don't change the reason/type/etc. we still need an update to happen here

status/condition_set_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 var _ = Describe("Conditions", func() {
 	It("should correctly toggle conditions", func() {
-		testObject := TestObject{}
+		testObject := &TestObject{ObjectMeta: metav1.ObjectMeta{Generation: 1}}
```
jonathan-innis, line 14: nit: may be easier to just call `testObject.Generation = 1`

jonathan-innis, line 14: Can we validate that an update that occurs that doesn't change the transition time still updates the observed generation on the condition type?

status/condition_set.go
wangamel: Synced offline, LastTransitionTime will not have to be updated along with observedGeneration. Updated.

status/condition_set_test.go
wangamel: Updated

wangamel: Unit test also updated to cover such case

status/condition_set.go
```diff
@@ -104,18 +104,27 @@ func (c ConditionSet) IsTrue(conditionTypes ...string) bool {
...
+	for _, cond := range c.object.GetConditions() {
+		if cond.Type != conditionType {
+			conditions = append(conditions, cond)
 		} else {
-			// If we'd only update the LastTransitionTime, then return.
-			condition.LastTransitionTime = c.LastTransitionTime
-			if reflect.DeepEqual(condition, c) {
-				return false
+			condition.LastTransitionTime = cond.LastTransitionTime
+			condition.ObservedGeneration = cond.ObservedGeneration
```
jonathan-innis, line 112: This won't update the observed generation when the condition is set for the first time. We should ideally update the observed generation always in the same spot that we were updating the time below so that we know that we always set it. There's probably some handling that we'll have to have so that we are still updating the ObservedGeneration but only update TransitionTime when there is a status update

jonathan-innis: I also have this feeling that there is a better way to factor this so that we don't have so many conditionals. Generally, prefer short-circuiting when you can if it's possible based on the behavior that we want here.

jonathan-innis: So, a couple things:
1. I still felt like we could factor this better so I tried my hand at it for a second. Does this work?

```go
func (c ConditionSet) Set(condition Condition) (modified bool) {
	conditionType := condition.Type
	var conditions []Condition
	for _, cond := range c.object.GetConditions() {
		if cond.Type != conditionType {
			conditions = append(conditions, cond)
		} else {
			condition.LastTransitionTime = metav1.Now()
			condition.ObservedGeneration = c.object.GetGeneration()
			if condition.Status == cond.Status {
				condition.LastTransitionTime = cond.LastTransitionTime
			}
			if reflect.DeepEqual(condition, cond) {
				return false
			}
		}
	}
	conditions = append(conditions, condition)
	// Sorted for convenience of the consumer, i.e. kubectl.
	sort.Slice(conditions, func(i, j int) bool { return conditions[i].Type < conditions[j].Type })
	c.object.SetConditions(conditions)

	// Recompute the root condition after setting any other condition
	c.recomputeRootCondition(conditionType)
	return true
}
```

2. I wanted to validate when the observedGeneration of the root condition should be updated. Is it set to the minimum of all dependent status conditions or is it set the whatever the current generation is? My gut would be that this would only be meaningful when it is set to the minimum -- that way you know that the current state of Readiness reflects an old view of state, but I need to check this in upstream.

wangamel:
```
if condition.Status == cond.Status {
				condition.LastTransitionTime = cond.LastTransitionTime
			}
```
It may be true/enough for our case, but the original comment, actually update the LastTransitionTime when any of Status/Reason/Message, so just checking Status changes the original logic how LastTransitionTime gets updated

jonathan-innis: This does the same, right? Basically, the check is: Did the status change? If it did, then the transition time should update. We shouldn't update the transition time if the reason updated, but we should still update the reason on the condition if there was a request to do so.

jonathan-innis: > I wanted to validate when the observedGeneration of the root condition should be updated

On this point: Thinking through it -- I think the correct behavior here is to update the root condition to Unknown if the observedGeneration doesn't match across. This is the known state of the root condition: We don't know whether it will be true or false since the dependent conditions haven't had the chance to catch-up yet. Once all of the status conditions match their observed generation to the latest, the root condition can update to a known state (true/false)

jonathan-innis: This also means that a constantly updating spec will never resolve into a "known" state for the Ready condition but I see this as expected since adding the observedGeneration to ready means that the Ready condition that we have should reflect the actual state at that generation. If we don't know it because we were in a state of transitioning, then we have to report that.

status/condition_set.go
```diff
@@ -104,18 +104,20 @@ func (c ConditionSet) IsTrue(conditionTypes ...string) bool {
...
+	for _, cond := range c.object.GetConditions() {
+		if cond.Type != conditionType {
+			conditions = append(conditions, cond)
 		} else {
-			// If we'd only update the LastTransitionTime, then return.
-			condition.LastTransitionTime = c.LastTransitionTime
-			if reflect.DeepEqual(condition, c) {
+			condition.LastTransitionTime = metav1.Now()
+			condition.ObservedGeneration = c.object.GetGeneration()
+			if condition.Status == cond.Status && cond.LastTransitionTime.UnixNano() > 0 {
```
jonathan-innis, line 113:
```suggestion
			if condition.Status == cond.Status && !cond.LastTransitionTime.IsZero() {
```

This is more standard

status/condition_set.go
```diff
@@ -209,9 +211,12 @@ func (c ConditionSet) recomputeRootCondition(conditionType string) {
 	} else {
 		c.Set(Condition{
 			Type: c.root,
-			// The root condition is no longer unknown as soon as any are false
+			// The root condition is no longer unknown as soon as any dependency condition false with latest observedGeneration
```
jonathan-innis, line 214:
```suggestion
			// The root condition is no longer unknown as soon as any dependent condition goes false with the latest observedGeneration
```

status/condition_set.go
```diff
@@ -209,9 +211,12 @@ func (c ConditionSet) recomputeRootCondition(conditionType string) {
...
 			Status: lo.Ternary(
-				lo.ContainsBy(conditions, func(condition Condition) bool { return condition.IsFalse() }),
+				lo.ContainsBy(conditions, func(condition Condition) bool {
+					return condition.IsFalse() &&
+						condition.ObservedGeneration == c.object.GetGeneration()
+				}),
 				metav1.ConditionFalse,
 				metav1.ConditionUnknown,
 			),
```
jonathan-innis, line 222: I think we may need some special handling to change the `condition.Status` to unknown for a sub-condition if the current status isn't from the latest generation. Otherwise, you're going to get some weird error messages that say that you have "UnhealthyDependents" but might actually have "True" status conditions for an old generation

jonathan-innis: I wonder if we can create a new Reason value `ReconcilingDependents` for cases when we are Unknown and we are still reconciling the state of other conditions

status/condition_set_test.go
```diff
@@ -64,9 +69,20 @@ var _ = Describe("Conditions", func() {
...
+		Expect(updatedBarCondition.ObservedGeneration).To(Equal(int64(1)))
 		// Dont transition if reason and message are the same
 		Expect(conditions.SetTrue(ConditionTypeFoo)).To(BeFalse())
 		Expect(conditions.SetFalse(ConditionTypeBar, "another-reason", "another-message")).To(BeFalse())
+		testObject.Generation = 2
+		Expect(conditions.SetFalse(ConditionTypeBar, "another-reason", "another-message")).To(BeTrue())
+		updatedBarCondition2 := conditions.Get(ConditionTypeBar)
+		Expect(updatedBarCondition2.LastTransitionTime.UnixNano()).To(BeNumerically("==", updatedBarCondition.LastTransitionTime.UnixNano()))
+		Expect(updatedBarCondition2.ObservedGeneration).To(Equal(int64(2)))
+		// root should be false when any dependency condition is false
```
jonathan-innis, line 81:
```suggestion
		// root should be false when any dependent condition is false
```

status/condition_set_test.go
```diff
@@ -64,9 +69,20 @@ var _ = Describe("Conditions", func() {
...
+		Expect(conditions.Root().GetStatus()).To(Equal(metav1.ConditionFalse))
+		// root should be unknown when no dependency condition is false and any observedGeneration doesn't match with latest generation
```
jonathan-innis, line 83:
```suggestion
		// root should be unknown when no dependent condition is false and any observedGeneration doesn't match with latest generation
```

status/condition_set.go
wangamel:
```
I think we may need some special handling to change the condition.Status to unknown for a sub-condition if the current status isn't from the latest generation. Otherwise, you're going to get some weird error messages that say that you have "UnhealthyDependents" but might actually have "True" status conditions for an old generation
```
 this situation is already set to Unknown from above logic?

jonathan-innis (approved):
LGTM 🚀

PR #89, opened 2024-11-06
feat: Create a Generic Status Controller 

engedaam:
*Issue #, if available:*

*Description of changes:*
- Use generalized Store for firing metrics based on status object
- Add a node status controller to monitor node status conditions  


status/constanst.go
```diff
@@ -0,0 +1,17 @@
+package status
```
jonathan-innis, line 1: This file is not named correctly -- I think that you meant to call it constants. Also, why not put this in the same file as the `metrics.go` since this applies to the metrics specifically?

status/node_controller.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/client-go/tools/record"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type node struct {
+	*corev1.Node
```
jonathan-innis, line 22: I don't think that you should make this specific to node -- ideally this works for any `client.Object` and we can just use the same conversion across the board to generalize this

status/node_controller.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (n node) GetConditions() []Condition {
+	return lo.Map(n.Status.Conditions, func(condition corev1.NodeCondition, _ int) Condition {
```
jonathan-innis, line 26: Rather than directly knowing what type of object this is, can we use unstructured to make this work generally for every type of objct?

status/store.go
```diff
@@ -0,0 +1,168 @@
...
+	"k8s.io/client-go/tools/record"
+)
+
+type Store struct {
+	eventRecorder      record.EventRecorder
+	observedConditions sync.Map // map[reconcile.Request]ConditionSet
+	terminatingObjects sync.Map // map[reconcile.Request]DeletionTimestamp
+}
+
+func NewStore(eventRecorder record.EventRecorder) Store {
```
jonathan-innis, line 24: Do we need this store implementation or can we wrap the standard status controller with a client.Object controller and just reuse its reconcile metric so we don't have to copy all of the logic and data store?

status/constanst.go
engedaam: Dropped the file

status/node_controller.go
engedaam: I'm now using unstructured with the `client.Object` type

status/store.go
engedaam: I have updated the PR to taken the changes

status/controller.go
```diff
@@ -65,41 +46,53 @@ func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
...
+func (c *GenericObjectController[T]) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		For(object.New[T]()).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
+		Named(fmt.Sprintf("operatorpkg.%s.status", strings.ToLower(reflect.TypeOf(object.New[T]()).Elem().Name()))).
+		Complete(c)
+}
+
+func (c *GenericObjectController[T]) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	unstrcut := lo.Must(NewUnstructuredAdapter(object.New[T]()))
```
jonathan-innis, line 80: I think this still contains more logic than my PR had -- is it needed? AFAIK, you can combine the Get logic in both of the controllers -- was there a gap in the changes that were proposed in #90?

status/unstructured_adapter.go
```diff
@@ -0,0 +1,49 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// unstructuredAdapter is an adapter for the status.Object interface. unstructuredAdapter
+// makes the assumption that status conditions are found on status.conditions path.
+type unstructuredAdapter struct {
+	*unstructured.Unstructured
+}
+
+func NewUnstructuredAdapter(obj client.Object) (*unstructuredAdapter, error) {
```
jonathan-innis, line 19: Why are we returning an error from this adapter? Beyond the `lo.Must`, it's not possible for this constructor to throw an error

status/unstrcutred_adapter_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var _ = Describe("Unstructured Adapter", func() {
+	It("Get unstructured conditions condition values values", func() {
+		testObject := &unstructured.Unstructured{Object: map[string]interface{}{
+			"status": map[string]interface{}{
+				"conditions": []interface{}{
+					map[string]interface{}{
```
jonathan-innis, line 17: Should we try multiple conditions here?

status/controller.go
engedaam: When I building out the testing for this including the Get call resulted in a `nil` pointer deference when trying to get an unstructured type with the adapted pointer. The client was not able to Get adapted unstructured pointer

jonathan-innis: > The client was not able to Get adapted unstructured pointer

Hmmm -- that doesn't make sense -- I was able to validate this by actually running it directly and this worked consistently for me.

jonathan-innis: I fixed this by changing the UnstructuredAdapter to compose an `unstructured.Unstructured` rather than composing an `*unstructured.Unstructured`. This line is the culprit (https://github.com/kubernetes-sigs/controller-runtime/blob/v0.19.1/pkg/client/fake/client.go#L541) -- which `nils` out everything in the object if you have any pointers -- which was causing the panic

status/controller_test.go
```diff
@@ -22,16 +23,23 @@ import (
...
+	kubeClient = fake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+	ctx = log.IntoContext(context.Background(), ginkgo.GinkgoLogr)
+})
+
 var _ = Describe("Controller", func() {
-	var ctx context.Context
-	var recorder *record.FakeRecorder
 	var controller *status.Controller[*TestObject]
-	var kubeClient client.Client
+	var generic_controller *status.GenericObjectController[*TestGenericObject]
```
jonathan-innis, line 38:
```suggestion
	var genericController *status.GenericObjectController[*TestGenericObject]
```

nit: Use go style -- no snake case in go style

status/controller_test.go
```diff
@@ -349,8 +702,9 @@ func GetMetric(name string, labels ...map[string]string) *prometheus.Metric {
 	return nil
 }
 
-func conditionLabels(t status.ConditionType, s metav1.ConditionStatus) map[string]string {
+func conditionLabels(k string, t status.ConditionType, s metav1.ConditionStatus) map[string]string {
```
jonathan-innis, line 705:
```suggestion
func conditionLabels(obj client.Object, t status.ConditionType, s metav1.ConditionStatus) map[string]string {
```

consider passing obj through this function over just the kind

status/controller_test.go
```diff
@@ -326,6 +336,349 @@ var _ = Describe("Controller", func() {
 			}()
 		}
 	})
+	It("should emit termination metrics when deletion timestamp is set", func() {
```
jonathan-innis, line 339: Should we create a separate context for the generic controller?

status/controller_test.go
engedaam: Updated to taken the `schema.GroupVersionKind` since that's all we need from the object

jonathan-innis (approved):
LGTM 🚀

PR #90, opened 2024-11-07
[DRAFT] Add generic object controller

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/status/adapter.go
+++ b/status/adapter.go
@@ -1,0 +1,49 @@
+package status
+
+import (
+	"encoding/json"
+
+	"github.com/awslabs/operatorpkg/object"
+	"github.com/samber/lo"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// UnstructuredAdapter is an adapter for the status.Object interface. unstructuredAdapter
+// makes the assumption that status conditions are found on status.conditions path.
+type UnstructuredAdapter struct {
+	*unstructured.Unstructured
+}
+
+func NewUnstructuredAdapter(obj client.Object) (*UnstructuredAdapter, error) {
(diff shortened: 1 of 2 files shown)


status/adapter.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// UnstructuredAdapter is an adapter for the status.Object interface. unstructuredAdapter
+// makes the assumption that status conditions are found on status.conditions path.
+type UnstructuredAdapter struct {
+	*unstructured.Unstructured
+}
+
+func NewUnstructuredAdapter(obj client.Object) (*UnstructuredAdapter, error) {
+	u := &unstructured.Unstructured{Object: lo.Must(runtime.DefaultUnstructuredConverter.ToUnstructured(obj))}
```
jonathan-innis, line 20: Checking the status conditions and validating that they exist on the object doesn't work as far as I can tell since conditions are not always present, meaning that we can really only validate it at runtime and assume that the developer correctly chooses objects that have status conditions on "status.conditions"

status/controller.go
```diff
@@ -207,6 +209,28 @@ func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (r
...
+	return &GenericObjectController[T]{
+		Controller: NewController[UnstructuredAdapter](client, eventRecorder),
+	}
+}
+
+func (c *GenericObjectController[T]) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	return c.reconcile(ctx, req, lo.Must(NewUnstructuredAdapter(object.New[T]())))
+}
+
+func (c *GenericObjectController[T]) Register(_ context.Context, m manager.Manager) error {
```
jonathan-innis, line 226: We can't drop this function unfortunately because it uses the `T` as the unstructured adapter if we don't write our own here -- which doesn't work for getting the name or generating the watch

jonathan-innis:
Closing in favor of #89

PR #91, opened 2024-11-08
feat: Add a events controller to fire Metric based on an event 

engedaam:
*Issue #, if available:*
- Metrics around the events supports better observability around state of operator objects outside of status conditions 

*Description of changes:*
- Adding a controller to watch `client.Objects` and emit a gauge based metrics on the count of an event for a Kubernetes object 
- Fire a metric for `operator_event_count` based on a object


events/controller.go
```diff
@@ -0,0 +1,63 @@
...
+			event := o.(*v1.Event)
+			return event.InvolvedObject.Kind == c.kind
+		}))).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
+		Named(fmt.Sprintf("operatorpkg.%s.events", strings.ToLower(c.kind))).
+		Complete(reconcile.AsReconciler(m.GetClient(), c))
+}
+
+func (c *Controller[T]) Reconcile(ctx context.Context, event *v1.Event) (reconcile.Result, error) {
+	// Deduplicate events on start-up
```
jonathan-innis, line 48: A longer description may be useful here -- I also think that this comment might be a little misleading -- it implies that we are doing some deduplication action, we aren't doing anything like that though, we are just making sure that we do't capture events that might have already occurred prior to the controller restart

events/controller.go
```diff
@@ -0,0 +1,63 @@
...
+	if c.startTime.Before(event.LastTimestamp.Time) {
+		group := strings.Split(event.InvolvedObject.APIVersion, "/")[0]
```
jonathan-innis, line 50: There's an upstream apiutil function parse the apiversion -- I think we should prefer that over doing the string parse ourselves

events/controller.go
```diff
@@ -0,0 +1,63 @@
...
+		EventCount.Inc(map[string]string{
+			MetricLabelGroup:        group,
+			MetricLabelKind:         event.InvolvedObject.Kind,
+			MetricLabelNamespace:    event.InvolvedObject.Namespace,
```
jonathan-innis, line 54: Let's drop namespace/name for now -- this is going to drastically increase our cardinality

events/controller.go
```diff
@@ -0,0 +1,63 @@
...
+			MetricLabelName:         event.InvolvedObject.Name,
+			MetricLabelEventType:    event.Type,
+			MetricLabelEventMessage: event.Message,
```
jonathan-innis, line 57: We absolutely can't do message and I question if we can even do reason -- we have to be _really_ careful with cardinality when it comes to counters -- because counters are going to exist for the lifetime of the process

events/controller.go
```diff
@@ -0,0 +1,63 @@
...
+			MetricLabelEventReason:  event.Reason,
+		})
+	}
+
+	return reconcile.Result{RequeueAfter: time.Second * 10}, nil
```
jonathan-innis, line 62: Why do we even need to requeue? This is going to lead to us double-counting -- we just need the create event

events/controller.go
```diff
@@ -0,0 +1,63 @@
...
+	return &Controller[T]{
+		kind:       reflect.TypeOf(object.New[T]()).Elem().Name(),
+		startTime:  clock.Now(),
+		kubeClient: client,
+	}
+}
+
+func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		For(&v1.Event{}, builder.WithPredicates(predicate.NewTypedPredicateFuncs(func(o client.Object) bool {
```
jonathan-innis, line 38: Something I don't know: Can events be updated? If so, how do we handle that so that we don't double-fire for events?

engedaam: Seems like an event can be updated/patched https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/#Event. An updated predicate should resolve that, no?

events/controller.go
engedaam: We need the reason, since that's what holds the context of why the event was fired

events/controller.go
jonathan-innis: Right -- I really don't think that it's likely that these events get updated/patched though -- it's probably fine to not handle this and only deal with this if it actually becomes a problem

events/controller.go
jonathan-innis: Reason might have an unknown cardinality since it's more freeform than type -- we should just see if we truly do need this

events/controller.go
```diff
@@ -0,0 +1,65 @@
...
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
+		Named(fmt.Sprintf("operatorpkg.%s.events", strings.ToLower(c.kind))).
+		Complete(reconcile.AsReconciler(m.GetClient(), c))
+}
+
+func (c *Controller[T]) Reconcile(ctx context.Context, event *v1.Event) (reconcile.Result, error) {
+	// We check if the event was created in the lifetime of this controller
+	// since we don't duplicate metrics on controller restart or lease handover
+	if c.startTime.Before(event.LastTimestamp.Time) {
+		groupVersion, err := schema.ParseGroupVersion(event.InvolvedObject.APIVersion)
```
jonathan-innis, line 52: `lo.Must`? Do we ever expect this to fail? So what's the value of requeueing?

events/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+	MetricSubsystem = "event"
+)
+
+// Cardinality is limited to # objects * # conditions
+var EventCount = pmetrics.NewPrometheusCounter(
+	metrics.Registry,
+	prometheus.CounterOpts{
+		Namespace: MetricNamespace,
+		Subsystem: MetricSubsystem,
+		Name:      "count",
```
jonathan-innis, line 27: The correct Prometheus name for a counter always ends with `total`, so something like `operator_node_event_count`

events/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+const (
+	MetricLabelGroup       = "group"
+	MetricLabelKind        = "kind"
+	MetricLabelEventType   = "type"
+	MetricLabelEventReason = "reason"
+)
+
+const (
+	MetricNamespace = "operator"
+	MetricSubsystem = "event"
```
jonathan-innis, line 18: Should this subsystem be prefixed with the object name? We already are thinking about doing this same thing for our status condition controller so should we do the same thing here?

events/controller.go
jonathan-innis: Discussed offline: Reason is actually the field that we care about for events, closing this comment since we'll keep both reason and type.

events/controller.go
engedaam: I think we can treat this as an invariant and expect the developer to have passed in an object with a valid APIVersion

events/controller.go
```diff
@@ -0,0 +1,76 @@
...
+			MetricLabelEventType:   event.Type,
+			MetricLabelEventReason: event.Reason,
+		})
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller[T]) Reset() {
+	c.eventCount.Reset()
```
jonathan-innis, line 75: Make eventCount public rather than building in a Reset() on the controller

events/controller.go
```diff
@@ -0,0 +1,76 @@
...
+		),
+	}
+}
+
+func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		For(&v1.Event{}, builder.WithPredicates(predicate.NewTypedPredicateFuncs(func(o client.Object) bool {
+			// Only reconcile on the object kind we care about
+			event := o.(*v1.Event)
+			return event.InvolvedObject.Kind == c.kind
```
jonathan-innis, line 52: Should we also check the group? What happens if there are two resources on the cluster that have the same kind and different groups?

events/suite_test.go
```diff
@@ -0,0 +1,179 @@
...
+	kubeClient = fake.NewClientBuilder().WithScheme(scheme.Scheme).WithIndex(&corev1.Event{}, "involvedObject.kind", func(o client.Object) []string {
+		evt := o.(*corev1.Event)
+		return []string{evt.InvolvedObject.Kind}
+	}).Build()
+	controller = events.NewController[*TestObject](kubeClient, fakeClock)
+	ctx = log.IntoContext(context.Background(), ginkgo.GinkgoLogr)
+})
+
+var (
+	SchemeBuilder = runtime.NewSchemeBuilder(func(scheme *runtime.Scheme) error {
```
jonathan-innis, line 53: Don't we already have a scheme builder elsewhere in our test objects directory?

events/suite_test.go
```diff
@@ -0,0 +1,179 @@
...
+		ExpectApplied(ctx, kubeClient, event)
+		_, err = reconcile.AsReconciler(kubeClient, controller).Reconcile(ctx, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(event)})
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(getMetric("operator_testobject_event_total", conditionLabels(corev1.EventTypeNormal, "reason")).GetCounter().GetValue()).To(BeEquivalentTo(1))
+	})
+})
+
+// +k8s:deepcopy-gen=true
+// +kubebuilder:object:root=true
```
jonathan-innis, line 105: Why do we need to copy over the DeepCopies? Don't we have a package for our test objects?

events/controller.go
```diff
@@ -0,0 +1,72 @@
...
+type Controller[T client.Object] struct {
+	objectGVK  schema.GroupVersionKind
+	startTime  time.Time
+	kubeClient client.Client
+	EventCount pmetrics.CounterMetric
+}
+
+func NewController[T client.Object](client client.Client, clock clock.Clock) *Controller[T] {
+	objectGVK := object.GVK(object.New[T]())
+	EventCountOpts.Subsystem = strings.ToLower(objectGVK.Kind)
```
rschalo, line 34: How does this not get overwritten and then shared when multiple event controllers are running?

events/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	MetricNamespace = "operator"
+)
+
+// Cardinality is limited to # objects * # conditions
+var EventCountOpts = prometheus.CounterOpts{
+	Namespace: MetricNamespace,
+	Name:      "event_total",
+	Help:      "The count of events of a given type for an object.",
+}
```
rschalo, line 23: Should this be a struct and then we have a function which accepts overrides and returns it?

events/suite_test.go
```diff
@@ -0,0 +1,138 @@
...
+			Expect(getMetric("operator_testobject_event_total", conditionLabels(fmt.Sprintf("Test-type-%d", i), fmt.Sprintf("Test-reason-%d", i)))).To(BeNil())
+		}
+
+		for i := range 5 {
+			_, err := reconcile.AsReconciler(kubeClient, controller).Reconcile(ctx, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(events[i])})
+			Expect(err).ToNot(HaveOccurred())
+		}
+
+		for i := range 5 {
+			Expect(getMetric("operator_testobject_event_total", conditionLabels(fmt.Sprintf("Test-type-%d", i), fmt.Sprintf("Test-reason-%d", i))).GetCounter().GetValue()).To(BeEquivalentTo(1))
```
rschalo, line 71: Can these not happen in the same loop?

events/controller.go
engedaam: These do get overridden each time we create a new controller, however since the counter is held in the controller struct and created with the object of that kind, we would always be tracking the right counter. Never the less I removed this favor of creating the metrics from a function

events/metrics.go
engedaam: I like that Idea! Updated the to use a common function

events/suite_test.go
engedaam: I wanted each set to happen one at a time. First creating all the events, reconciling all the values and then validating each one. The idea being that we can check if multiple  event reconciliation was happening. I'm committed to that idea, so I can take in your comment

rschalo: If you'd like to test multiple event reconciliation, would a waitgroup make sense? https://github.com/aws/karpenter-provider-aws/blob/7e8269bc5b333afdd6863d9e7b933a5bc049cd2f/test/suites/scale/deprovisioning_test.go#L183

events/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+	MetricLabelKind        = "kind"
+	MetricLabelEventType   = "type"
+	MetricLabelEventReason = "reason"
+)
+
+const (
+	MetricNamespace = "operator"
+)
+
+func event_count_metric(object_name string) pmetrics.CounterMetric {
```
rschalo, line 20: nit: `counterMetric`

events/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+	MetricLabelKind        = "kind"
+	MetricLabelEventType   = "type"
+	MetricLabelEventReason = "reason"
+)
+
+const (
+	MetricNamespace = "operator"
+)
+
+func counterMetric(object_name string) pmetrics.CounterMetric {
```
rschalo, line 20:
```suggestion
func counterMetric(objectName string) pmetrics.CounterMetric {
```

events/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+	return pmetrics.NewPrometheusCounter(
+		metrics.Registry,
+		prometheus.CounterOpts{
+			Namespace: MetricNamespace,
+			Subsystem: object_name,
```
rschalo, line 25:
```suggestion
			Subsystem: objectName,
```

events/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	pmetrics "github.com/awslabs/operatorpkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	MetricLabelEventType   = "type"
+	MetricLabelEventReason = "reason"
```
rschalo, line 11:
```suggestion
	MetricLabelType   = "type"
	MetricLabelReason = "reason"
```
Pretty sure it be just this because it's scoped to the `events` pkg already, right?

engedaam: Here I was following the convention we had set out in the status condition controller https://github.com/awslabs/operatorpkg/blob/main/status/metrics.go#L14-L16.

engedaam: I can go a head and make this common if that's the approach we want to take

events/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func counterMetric(objectName string) pmetrics.CounterMetric {
+	return pmetrics.NewPrometheusCounter(
+		metrics.Registry,
+		prometheus.CounterOpts{
+			Namespace: pmetrics.Namespace,
+			Subsystem: objectName,
+			Name:      "event_total",
+			Help:      "The count of events of a given type for an object.",
```
jonathan-innis, line 21:
```suggestion
			Help:      "The total count of events of a given type for an object.",
```

test/object.go
```diff
@@ -40,3 +43,91 @@ func RandomName() string {
...
+var (
+	SchemeBuilder = runtime.NewSchemeBuilder(func(scheme *runtime.Scheme) error {
+		scheme.AddKnownTypes(schema.GroupVersion{Group: APIGroup, Version: "v1alpha1"}, &TestObject{})
+		return nil
+	})
+)
+
+// +k8s:deepcopy-gen=true
+// +kubebuilder:object:root=true
+type TestObject struct {
```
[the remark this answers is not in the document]
engedaam: I updated this be `CustomObject`. This should now read `test.CustomObject{}`, which should be clear that it's a testing object that implements `Object`

PR #92, opened 2024-11-08
feat: Add interface functions for prometheus metrics to extend to other metric types

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

This change wraps the prometheus metric using an interface so that we can build multiple metric paths into the metrics that we publish. By default, we will route the metrics to prometheus but since the variables are exposed externally, users can modify the metric to create their own metric pipeline.

For instance

```
status.TerminationDuration = pmetrics.NewMultiObservation(
	status.TerminationDuration,
	pmetrics.NewEMFObservation(writer, "operatorpkg", "operator_termination_duration_seconds", []string{"group", "kind", "namespace"}),
)
```

### DIFF
--- a/metrics/emf.go
+++ b/metrics/emf.go
@@ -1,0 +1,189 @@
+package metrics
+
+import (
+	"encoding/json"
+	"io"
+	"time"
+
+	"github.com/samber/lo"
+)
+
+type EMF struct {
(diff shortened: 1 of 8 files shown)


njtran (approved):
LGTM

Issue #87, opened 2024-10-28
Status Conditions Metrics should be per-object type 


njtran:
Status Conditions currently include the type of the object as a dimension of the metric. Metric aggregations are usually done on a metric name and aggregated based on the dimensions of the metric. If I wanted to say get the number of nodes for label `team: app-a` that have a MemoryPressure status condition, I can't do it with the given status conditions. I'd have to cross reference each of the status conditions, with a metric on the node for labels, and then aggregate them, which has odd aggregation behavior as there's not a 1-1 for each metric, having to do joins.

Ideally, we can just auto-gen metrics in the `operator_status` system and add a subsystem with the type of the object we're emitting for, and do the usual status condition metrics with specific labels we care about added in.

jonathan-innis:
Closed by #98

PR #98, opened 2024-11-13
feat: Add Object Specific Metric for Status Condition Controller

engedaam:
Issue https://github.com/awslabs/operatorpkg/issues/87

*Description of changes:*

Auto-gen metric names based on the kind that is passed into the controller. The system for the metric will be `operator` and subsystem is `<object kind>_status_condition`. The labels used are the same statically defined metrics, with exception of the kind label, that was removed. These new metrics can be enabled using the controller options passed at controller creation. The new auto generated metrics will look as such:
- operator_<object kind>_status_condition_transitions_total
- operator_<object kind>_status_condition_transition_seconds
- operator_<object kind>_status_condition_current_status_seconds
- operator_<object kind>_status_condition_count
- operator_<object kind>_termination_current_time_seconds
- operator_<object kind>_termination_duration_seconds

### DIFF
--- a/status/controller.go
+++ b/status/controller.go
@@ -9,9 +9,12 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	pmetrics "github.com/awslabs/operatorpkg/metrics"
 	"github.com/awslabs/operatorpkg/object"
+	"github.com/awslabs/operatorpkg/option"
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


status/controller.go
```diff
@@ -23,24 +25,52 @@ import (
...
+	ConditionDuration             pmetrics.ObservationMetric
+	ConditionCount                pmetrics.GaugeMetric
+	ConditionCurrentStatusSeconds pmetrics.GaugeMetric
+	ConditionTransitionsTotal     pmetrics.CounterMetric
+	TerminationCurrentTimeSeconds pmetrics.GaugeMetric
+	TerminationDuration           pmetrics.ObservationMetric
 }
 
-func NewController[T Object](client client.Client, eventRecorder record.EventRecorder) *Controller[T] {
+type ControllerOpts struct {
```
engedaam, line 41: We need to be able to pass in the type for client.object, since unstructured types do not have a Kind.

status/controller.go
```diff
@@ -23,24 +25,52 @@ import (
...
+
+func NewController[T Object](client client.Client, eventRecorder record.EventRecorder, opts ...ControllerOpts) *Controller[T] {
+	var gvk schema.GroupVersionKind
+	switch len(opts) {
+	case 0:
+		gvk = object.GVK(object.New[T]())
+	case 1:
+		gvk = opts[0].GVK
+	default:
+		panic("expected no or one argument for controller options")
```
engedaam, line 53: In my mind, I could not think of a case where we would need to pass in more then one controller option

status/metrics.go
```diff
@@ -39,6 +41,23 @@ var ConditionDuration = pmetrics.NewPrometheusHistogram(
...
+	return pmetrics.NewPrometheusHistogram(
+		metrics.Registry,
+		prometheus.HistogramOpts{
+			Namespace: MetricNamespace,
+			Subsystem: fmt.Sprintf("%s_%s", objectName, MetricSubsystem),
+			Name:      "transition_seconds",
+			Help:      "The amount of time a condition was in a given state before transitioning. e.g. Alarm := P99(Updated=False) > 5 minutes",
+		},
+		[]string{
+			MetricLabelGroup,
```
jonathan-innis, line 54: You aren't going to be able to register this metric twice so I think that we should drop group from here. Same is true of all of the other metrics that we are generating in this file

status/controller.go
```diff
@@ -212,3 +222,33 @@ func (c *Controller[T]) reconcile(ctx context.Context, req reconcile.Request, o
...
+
+func (c *Controller[T]) incCounterMetric(current pmetrics.CounterMetric, deprecated pmetrics.CounterMetric, labels map[string]string) {
+	current.Inc(labels)
+	labels[MetricLabelKind] = c.gvk.Kind
+	deprecated.Inc(labels)
+}
+
+func (c *Controller[T]) setGaugeMetric(current pmetrics.GaugeMetric, deprecated pmetrics.GaugeMetric, value float64, labels map[string]string) {
+	current.Set(value, labels)
+	labels[MetricLabelKind] = c.gvk.Kind
```
jonathan-innis, line 234: Nice!

status/unstructured_adapter_test.go
```diff
@@ -35,7 +35,7 @@ var _ = Describe("Unstructured Adapter", func() {
 			Kind:    "testKind",
 		})
 
-		conditionObj := status.NewUnstructuredAdapter(testObject)
+		conditionObj := status.NewUnstructuredAdapter[*TestObject](testObject)
```
jonathan-innis, line 38: musing: Don't do it in this PR but we may be able to get rid of passing the object into the adapter entirely now that the adapter itself carries the generic type

status/controller_test.go
```diff
@@ -20,23 +20,20 @@ import (
...
-	ctx = log.IntoContext(context.Background(), ginkgo.GinkgoLogr)
-
-	status.ConditionDuration.Reset()
-	status.ConditionCount.Reset()
-	status.ConditionCurrentStatusSeconds.Reset()
-	status.ConditionTransitionsTotal.Reset()
-	status.TerminationCurrentTimeSeconds.Reset()
-	status.TerminationDuration.Reset()
+var _ = AfterEach(func() {
+	status.ConditionDuration.Unregister(metrics.Registry)
```
jonathan-innis, line 31: You don't need to expose this as a method on the interface to get this to work. You can achieve the same by type-casting this into a prometheus metric like `status.ConditionDuration.(*metrics.PrometheusHistogram).Unregister(metrics.Registry)`

status/controller.go
```diff
@@ -24,24 +27,55 @@ import (
...
+	TerminationDuration           pmetrics.ObservationMetric
 }
 
-func NewController[T Object](client client.Client, eventRecorder record.EventRecorder) *Controller[T] {
+type ControllerOpts struct {
+	EnableObjectBasedMetrics bool
+}
+
+func NewController[T Object](client client.Client, eventRecorder record.EventRecorder, opts ...ControllerOpts) *Controller[T] {
+	options := ControllerOpts{}
```
jonathan-innis, line 49: There's an options package in `operatorpkg`. Should we use that?

status/controller.go
```diff
@@ -24,24 +27,55 @@ import (
...
+	ConditionCount                pmetrics.GaugeMetric
+	ConditionCurrentStatusSeconds pmetrics.GaugeMetric
+	ConditionTransitionsTotal     pmetrics.CounterMetric
+	TerminationCurrentTimeSeconds pmetrics.GaugeMetric
+	TerminationDuration           pmetrics.ObservationMetric
 }
 
-func NewController[T Object](client client.Client, eventRecorder record.EventRecorder) *Controller[T] {
+type ControllerOpts struct {
+	EnableObjectBasedMetrics bool
```
jonathan-innis, line 45: IMO, we should the object-prefixed by default and then we should allow you to opt-in to the other metric type where we use group/kind labels to differentiate the types. That's the "deprecated" way so we should be trying to get folks off of that as quickly as we can

status/metrics.go
```diff
@@ -18,108 +20,217 @@ const (
 )
 
 // Cardinality is limited to # objects * # conditions * # objectives
-var ConditionDuration = pmetrics.NewPrometheusHistogram(
-	metrics.Registry,
-	prometheus.HistogramOpts{
+var ConditionDuration pmetrics.ObservationMetric
+
+func conditionDurationMetric(objectName string, enableObjectBasedMetric bool) pmetrics.ObservationMetric {
```
jonathan-innis, line 25: To me -- this seems kind of a weird way to handle creating the ConditionDuration metric -- this is only going to be assigned the first time this function is used -- which isn't really what you want for globals -- you'd want them to be instantiated as soon as the package is imported so you don't accidentally have folks importing and using `nil` values for things that they wouldn't expect to be `nil`

status/controller_test.go
engedaam: Oh wow, that will simplify this a lot, thanks!

jonathan-innis (approved):
LGTM 🚀

PR #106, opened 2024-11-25
chore: Consider an absent condition to be Unknown

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

This change updates our condition logic so that we consider a lack of a condition as a condition in an `Unknown` status. This also ensures that the first time that we set the StatusCondition that it's set to the CreationTimestamp since the first time that we set the condition, the condition hasn't transitioned so the "LastTransitionTime" of this condition was actually when the object was created


object/object.go
```diff
@@ -56,6 +56,7 @@ func Unmarshal[T any](raw []byte) *T {
 
 func Hash(o any) string {
 	h := fnv.New64a()
-	hash.DeepHashObject(h, o)
-	return strconv.FormatUint(uint64(h.Sum64()), 10)
+	h.Reset()
```
jonathan-innis, line 59: This was needed to drop the Kubernetes dependency -- we shouldn't be relying on this repo as a dependency

status/condition_set.go
```diff
@@ -54,9 +54,16 @@ type ConditionSet struct {
...
+	// Set the root condition first to get consistent timing for LastTransitionTime
+	for _, t := range append([]string{r.root}, r.dependents...) {
 		if cs.Get(t) == nil {
-			cs.SetUnknown(t)
+			cs.Set(Condition{
+				Type:               t,
+				Status:             metav1.ConditionUnknown,
+				Reason:             "AwaitingReconciliation",
+				Message:            "object is awaiting reconciliation",
+				LastTransitionTime: object.GetCreationTimestamp(),
```
ellistarn, line 65: Optional: just call SetUnknown, and make it set creation timestamp if not currently set.

status/condition_set.go
```diff
@@ -99,12 +106,19 @@ func (c ConditionSet) IsTrue(conditionTypes ...string) bool {
 	return true
 }
 
+func (c ConditionSet) IsNormalCondition(t string) bool {
```
ellistarn, line 109: I really liked the dependent/independent terms you mentioned the other day.
```suggestion
func (c ConditionSet) IsDependentCondition(t string) bool {
```

status/condition_set.go
```diff
@@ -99,12 +106,19 @@ func (c ConditionSet) IsTrue(conditionTypes ...string) bool {
...
 // Set sets or updates the Condition on Conditions for Condition.Type.
 // If there is an update, Conditions are stored back sorted.
 func (c ConditionSet) Set(condition Condition) (modified bool) {
 	conditionType := condition.Type
 	var conditions []Condition
-	condition.LastTransitionTime = metav1.Now()
+	// If we get passed a LastTransitionTime, we should respect it
+	if condition.LastTransitionTime.IsZero() {
+		condition.LastTransitionTime = metav1.Now()
+	}
```
ellistarn, line 121: Can't we detect it's not set in here and set w/ creation timestamp?

status/controller.go
```diff
@@ -225,17 +225,27 @@ func (c *Controller[T]) reconcile(ctx context.Context, req reconcile.Request, o
 			MetricLabelConditionStatus: string(condition.Status),
 			pmetrics.LabelReason:       condition.Reason,
 		})
+		var duration time.Duration
+		var status metav1.ConditionStatus
 		if observedCondition == nil {
```
ellistarn, line 230: I wonder if there's a pattern here where it's never nil? We can guarantee that it's initialized whenever we try to retrieve a known condition. I'd need to touch the code to find the simplest path, but I suspect there's a simpler pattern than everything you're needing to do here.

test/object.go
```diff
@@ -33,6 +33,7 @@ func Object[T client.Object](base T, overrides ...T) T {
 	for _, src := range append([]T{base}, overrides...) {
 		lo.Must0(mergo.Merge(dest, src, mergo.WithOverride))
 	}
+	dest.SetCreationTimestamp(metav1.Now())
```
ellistarn, line 36: This should be set when applied by the API Server

status/condition_set.go
jonathan-innis: > and make it set creation timestamp if not currently set

How do you do that? You don't access to the creation timestamp since SetUnknown is on the condition set. You need some way to pass-down that detail

jonathan-innis: Ah, didn't realize that the ConditionSet had access to the object -- should have actually looked at the struct

status/condition_set.go
jonathan-innis: Didn't realize that we had access to the root object

test/object.go
jonathan-innis: It's not -- I tried it and I think because we are using a mock it doesn't get instantiated.

status/controller.go
jonathan-innis: > I wonder if there's a pattern here where it's never nil

It's not clear to me how that would work -- these conditions are stored so it's not clear to me how we could guarantee that the map is hydrated before we decide to do any kind of computation

jonathan-innis: > but I suspect there's a simpler pattern than everything you're needing to do here

Yep, you're right -- I'm not quite sure what I was thinking but I cleaned it up now 😄

object/object.go
jonathan-innis: It mostly just broke GoLand for dependency resolution

ellistarn (approved):

PR #112, opened 2024-12-04
fix: Store object in map on object termination and fix resource leak

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

Store object in map during object termination. Because we were not storing the older version of the object, we were not able to get additional properties from the object during object termination. This fixes the code so that we store the object and can grab additional properties.

Note: This also fixes a resource leak that was introduced with the terminatingObjects map

### DIFF
--- a/status/controller.go
+++ b/status/controller.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rschalo (approved):
lgtm

PR #115, opened 2024-12-18
Updating ExpectDelete to ignores already deleted objects

sarthakkothari:
*Description of changes:*
Updating ExpectDelete to ignores already deleted objects

### DIFF
--- a/test/expectations/expectations.go
+++ b/test/expectations/expectations.go
@@ -162,8 +162,8 @@ func ExpectStatusUpdated(ctx context.Context, c client.Client, objects ...client
 func ExpectDeleted(ctx context.Context, c client.Client, objects ...client.Object) {
 	GinkgoHelper()
 	for _, o := range objects {
-		Expect(c.Delete(ctx, o)).To(Succeed())
-		Expect(c.Get(ctx, client.ObjectKeyFromObject(o), o)).To(Or(Succeed(), MatchError(ContainSubstring("not found"))))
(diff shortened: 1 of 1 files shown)


chu-yik (approved):

test/expectations/expectations.go
```diff
@@ -167,6 +167,14 @@ func ExpectDeleted(ctx context.Context, c client.Client, objects ...client.Objec
 	}
 }
 
+func ExpectAlreadyDeletedOrDelete(ctx context.Context, c client.Client, objects ...client.Object) {
+	GinkgoHelper()
+	for _, o := range objects {
+		Expect(c.Delete(ctx, o)).To(Or(Succeed(), MatchError(ContainSubstring("not found"))))
+		Expect(c.Get(ctx, client.ObjectKeyFromObject(o), o)).To(Or(Succeed(), MatchError(ContainSubstring("not found"))))
+	}
```
njtran, line 175: Any reason why we wouldn't want this to be the default behavior of ExpectDeleted()? I also think it'd be nice if we could use the controller runtime client.IgnoreNotFound() here

sarthakkothari: >  why we wouldn't want this to be the default behavior ExpectDeleted()
I dont know if some packages are using the function in the old context where they want the function to imply that object existed and only then delete.

njtran (approved):

PR #121, opened 2025-01-18
chore: Match ExpectApplied expectation implementation with implementation from karpenter core

jeevanpuchakay:
*Issue #, if available:* https://github.com/kubernetes-sigs/karpenter/issues/1348

*Description of changes:* 
1. Match `ExpectApplied` expectation implementation with implementation from karpenter core
2. Current implementation creates object in #L97 but does not update it with the attributes that were originally passed to `ExpectApplied`.
3. Reference - https://github.com/jeevanpuchakay/karpenter/blob/ff44f7325bf006fecf66e2a5684a7d8700e51cfd/pkg/test/expectations/expectations.go#L151-L181
4. Currently there are no consumers to `ExpectApplied` method in operatorpkg.
5. Context - This is part of a larger effort to consolidate all expectations in `operatorpkg` and use them in karpenter (core) package. 

### DIFF
--- a/test/expectations/expectations.go
+++ b/test/expectations/expectations.go
@@ -89,6 +89,9 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
 	GinkgoHelper()
 	for _, o := range objects {
 		current := o.DeepCopyObject().(client.Object)
+		statuscopy := o.DeepCopyObject().(client.Object) // Snapshot the status, since create/update may override
+		deletionTimestampSet := !o.GetDeletionTimestamp().IsZero()
+
 		// Create or Update
 		if err := c.Get(ctx, client.ObjectKeyFromObject(current), current); err != nil {
 			if errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


test/expectations/expectations.go
```diff
@@ -89,6 +89,9 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
 	GinkgoHelper()
 	for _, o := range objects {
 		current := o.DeepCopyObject().(client.Object)
+		statuscopy := o.DeepCopyObject().(client.Object) // Snapshot the status, since create/update may override
+		deletionTimestampSet := !o.GetDeletionTimestamp().IsZero()
```
jonathan-innis, line 93: I'm not sure that I'm convinced by the deletionTimestamp logic that's present here -- this might have been in the karpenter implementation, but I'm not sure that this makes sense and is something that we want to pull in here

jonathan-innis:
@jeevanpuchakay Looks like this PR has gone a bit stale. If we don't hear from you over the next couple weeks we'll move this to closed.

PR #124, opened 2025-01-24
chore: Avoid Event List call by using Clinet go Watcher

engedaam:
*Issue N/A

*Description of changes:*
- For cluster with large event volume size, we during controller-runtime cache hydration the initial list call made to the APIServer has heavy load on the API Server. Since we don't care about events before controller startup,  we will remove the list call done by controller-runtime and setting watcher directly on event objects.

### DIFF
--- a/events/controller.go
+++ b/events/controller.go
@@ -8,15 +8,17 @@ import (
 
 	pmetrics "github.com/awslabs/operatorpkg/metrics"
 	"github.com/awslabs/operatorpkg/object"
+	"github.com/awslabs/operatorpkg/singleton"
+	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/clock"
 	controllerruntime "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -25,31 +27,32 @@ type Controller[T client.Object] struct {
 	startTime  time.Time
 	kubeClient client.Client
 	EventCount pmetrics.CounterMetric
+	eventWatch watch.Interface
 }
 
-func NewController[T client.Object](client client.Client, clock clock.Clock) *Controller[T] {
+func NewController[T client.Object](ctx context.Context, client client.Client, clock clock.Clock, kubernetesInterface kubernetes.Interface) *Controller[T] {
 	gvk := object.GVK(object.New[T]())
 	return &Controller[T]{
 		gvk:        gvk,
 		startTime:  clock.Now(),
 		kubeClient: client,
 		EventCount: eventTotalMetric(strings.ToLower(gvk.Kind)),
+		eventWatch: lo.Must(kubernetesInterface.CoreV1().Events("").Watch(ctx, metav1.ListOptions{})),
 	}
 }
 
-func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
+func (c *Controller[T]) Register(ctx context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
-		For(&v1.Event{}, builder.WithPredicates(predicate.NewTypedPredicateFuncs(func(o client.Object) bool {
-			// Only reconcile on the object kind we care about
-			event := o.(*v1.Event)
-			return event.InvolvedObject.Kind == c.gvk.Kind && event.InvolvedObject.APIVersion == c.gvk.GroupVersion().String()
-		}))).
-		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
 		Named(fmt.Sprintf("operatorpkg.%s.events", strings.ToLower(c.gvk.Kind))).
-		Complete(reconcile.AsReconciler(m.GetClient(), c))
+		WatchesRawSource(singleton.ChannelSource[*v1.Event](ctx, c.eventWatch.ResultChan())).
+		Complete(reconcile.AsReconciler(c.kubeClient, c))
 }
 
 func (c *Controller[T]) Reconcile(ctx context.Context, event *v1.Event) (reconcile.Result, error) {
+	if event.InvolvedObject.Kind != c.gvk.Kind && event.InvolvedObject.APIVersion != c.gvk.GroupVersion().String() {
+		return reconcile.Result{}, nil
+	}
+
 	// We check if the event was created in the lifetime of this controller
 	// since we don't duplicate metrics on controller restart or lease handover
 	if c.startTime.Before(event.LastTimestamp.Time) {
--- a/events/controller.go
+++ b/events/controller.go
@@ -49,7 +49,7 @@ func (c *Controller[T]) Register(ctx context.Context, m manager.Manager) error {
 }
 
 func (c *Controller[T]) Reconcile(ctx context.Context, event *v1.Event) (reconcile.Result, error) {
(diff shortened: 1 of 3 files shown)


events/controller.go
```diff
@@ -25,31 +27,32 @@ type Controller[T client.Object] struct {
...
 	return controllerruntime.NewControllerManagedBy(m).
-		For(&v1.Event{}, builder.WithPredicates(predicate.NewTypedPredicateFuncs(func(o client.Object) bool {
-			// Only reconcile on the object kind we care about
-			event := o.(*v1.Event)
-			return event.InvolvedObject.Kind == c.gvk.Kind && event.InvolvedObject.APIVersion == c.gvk.GroupVersion().String()
-		}))).
-		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
 		Named(fmt.Sprintf("operatorpkg.%s.events", strings.ToLower(c.gvk.Kind))).
-		Complete(reconcile.AsReconciler(m.GetClient(), c))
+		WatchesRawSource(singleton.ChannelSource[*v1.Event](ctx, c.eventWatch.ResultChan())).
```
jonathan-innis, line 47: Are you still able to use a predicate here to filter out things that we put into the queue so we don't end up with a bunch of things that are in the queue that we won't end up processing anyways?

singleton/controller.go
```diff
@@ -36,3 +38,33 @@ func Source() source.Source {
 		},
 	})
 }
+
+func ChannelSource[T client.Object](ctx context.Context, objectChan <-chan watch.Event) source.Source {
+	eventSource := make(chan event.GenericEvent, 1)
```
jonathan-innis, line 43: Should this channel buffer be larger? The writer is going to be forced to wait if the writer can't keep up so I wonder if we need to keep a larger channel buffer to make sure that we don't get stuck (it's pretty cheap from a memory perspective and should prevent our process from hanging on writes (maybe 1000?)

singleton/controller.go
```diff
@@ -36,3 +38,33 @@ func Source() source.Source {
...
+				eventSource <- event.GenericEvent{
+					Object: delta.Object.(T),
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}(ctx, objectChan, eventSource)
+
+	return source.Channel(eventSource, handler.Funcs{
```
jonathan-innis, line 63: Why can't we read directly from the watch channel here?

events/suite_test.go
```diff
@@ -46,13 +48,18 @@ func Test(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
+	ctx = context.Background()
+	ctx = log.IntoContext(context.Background(), ginkgo.GinkgoLogr)
```
rschalo, line 52: Did you mean to create a new background context here?

engedaam: Cleaning that up with a few updates to the tests

singleton/controller.go
engedaam: Yeah, that seems reasonable

singleton/controller.go
engedaam: We are now using the watch channel directly

singleton/controller.go
```diff
@@ -27,6 +31,39 @@ func AsReconciler(reconciler Reconciler) reconcile.Reconciler {
...
+func AsChannelObjectReconciler[T client.Object](watchEvents <-chan watch.Event, reconciler ChannalObjectReconciler[T]) reconcile.Reconciler {
+	return reconcile.Func(func(ctx context.Context, r reconcile.Request) (reconcile.Result, error) {
+		var errs error
+		var results []reconcile.Result
+		e := <-watchEvents
+		if e.Object == nil {
+			return reconcile.Result{RequeueAfter: RequeueImmediately}, nil
+		}
+		res, err := reconciler.Reconcile(ctx, e.Object.(T))
+		errs = multierr.Append(errs, err)
```
jonathan-innis, line 47: Why use a multierr here?

singleton/controller.go
```diff
@@ -27,6 +31,39 @@ func AsReconciler(reconciler Reconciler) reconcile.Reconciler {
...
+		results = append(results, res)
```
jonathan-innis, line 48: Why use a multi-result here? You're just going through one object with one result/error combo

singleton/controller.go
```diff
@@ -27,6 +31,39 @@ func AsReconciler(reconciler Reconciler) reconcile.Reconciler {
 	})
 }
 
+type ChannalObjectReconciler[T client.Object] interface {
```
jonathan-innis, line 34: This is so bespoke and specific to events, I'd honestly just apply this to the event controller alone and call it good. This is basically shimming a watch informer on top of controller-runtime. You could do this same thing with just creating a watch informer and firing a metric when you get an Create event on an event, but we're basically just wrapping this in controller-runtime's interface so that it adheres to the same interface for all controllers.

Unless n > 2, I wouldn't generalize

singleton/controller.go
engedaam: The handling here was meant to be extendable for multiple reconciliation for multiple objects pushed to the watch channel. Dropping this whole section as it will not be necessary, since I will be handling taking this comment above: https://github.com/awslabs/operatorpkg/pull/124/files#r1930026657

singleton/controller.go
engedaam: > This is basically shimming a watch informer on top of controller-runtime.

Yeah, this was the only purpose for the reconciler.  I thought this might be useful as a reconciler in the future, but can drop it for now.

jonathan-innis:
Small nit just about commit cleanliness: You should maybe consider squashing commits so you don't end-up with a 22-commit PR -- doesn't really matter since we end-up squashing commits in the PR anyways, but something to note for other repos that might use rebase merges

events/controller.go
```diff
@@ -8,48 +8,50 @@ import (
...
-	EventCount pmetrics.CounterMetric
+	gvk               schema.GroupVersionKind
+	startTime         time.Time
+	kubeClient        client.Client
+	EventCount        pmetrics.CounterMetric
+	EventWatchChannel <-chan watch.Event
 }
 
-func NewController[T client.Object](client client.Client, clock clock.Clock) *Controller[T] {
+func NewController[T client.Object](client client.Client, clock clock.Clock, channel <-chan watch.Event) *Controller[T] {
```
jonathan-innis, line 30: Can you build this in a way that you don't require the caller to start the watch themselves and have to implement the channel?

events/suite_test.go
```diff
@@ -46,13 +50,18 @@ func Test(t *testing.T) {
...
 		evt := o.(*corev1.Event)
 		return []string{evt.InvolvedObject.Kind}
 	}).Build()
-	controller = events.NewController[*test.CustomObject](kubeClient, fakeClock)
-	ctx = log.IntoContext(context.Background(), ginkgo.GinkgoLogr)
+
+	client := fake.NewSimpleClientset()
+	eventChannel = make(chan watch.Event, 1000)
+	controller = events.NewController[*test.CustomObject](ctx, kubeClient, fakeClock, client)
+	controller.EventWatchChannel = eventChannel
```
jonathan-innis, line 64: Nice!

events/controller.go
```diff
@@ -8,48 +8,57 @@ import (
...
-		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
 		Named(fmt.Sprintf("operatorpkg.%s.events", strings.ToLower(c.gvk.Kind))).
-		Complete(reconcile.AsReconciler(m.GetClient(), c))
+		WatchesRawSource(singleton.Source()).
+		Complete(singleton.AsReconciler(c))
 }
 
-func (c *Controller[T]) Reconcile(ctx context.Context, event *v1.Event) (reconcile.Result, error) {
+func (c *Controller[T]) Reconcile(ctx context.Context) (reconcile.Result, error) {
+	e := <-c.EventWatchChannel
```
jonathan-innis, line 56: I just took a look at the underlying code and the implementation of the ResultChan uses an unbuffered channel -- that means new writes are going to get blocked on reads here -- I'm concerned that we may block the watch stream if we don't either pull these events into a buffered channel OR use multiple goroutines to process these events. 

I'm trying to think of a good way to handle this while keeping the same interface so that we keep the same logging behavior and metrics -- nothing comes to mind immediately, but I think it's something that we should explore. Minimally, we can kick off a goroutine in the register that pulls data off of the watch channel and puts it in a separate buffered channel -- another option is to keep a static cache of events as we see them, enqueue a reconcile.Request and then read from that local cache (this is basically exactly what controller-runtime does with a read cache without the deletions)

jonathan-innis (dismissed):
LGTM 🚀

PR #138, opened 2025-03-19
Bump all obersved generation version of all conditions when object is deleting

sarthakkothari:
…

*Issue #, if available:*

*Description of changes:*

If a status condition is changed to true while the object is deleting, the root condition will transition to unknown as all the dependent conditions as from previous generation. We bump the generation of all the conditions so that the root condition remains the same as it was.


jonathan-innis (approved):
LGTM 🚀

PR #142, opened 2025-04-14
chore: Add the structured error package (`serrors`)

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

### `serrors.Wrap`

This change adds the `serrors` package which allows callers to create structured errors with key/value pairs for better structured logging. Wrapping errors with `serrors.Wrap` will add values. If the error is wrapped multiple times, key/value pairs will be pulled out when the error is unwrapped during logging

```
for _, val := range req.Values {
	it, err := c.getInstanceType(val)
	if err != nil {
		return nil, serrors.Wrap(fmt.Errorf("instance type not found"), "instance-type", val)
	}
        ...
}
```

```
{"level":"ERROR","time":"2025-04-14T06:49:10.528Z","logger":"controller","caller":"cloudprovider/cloudprovider.go:81","message":"failed launching nodeclaim","commit":"8da13bb","controller":"nodeclaim.lifecycle","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"default-ghcsc"},"namespace":"","name":"default-ghcsc","reconcileID":"478e0d8b-609e-4ce1-b4dd-2c2eb03eae06","instance-type":"m5.large","error":"instance type not found"}
```

### `serrors.Logger`

This change wraps a standard logger with a `serrors.Logger`. This logger unpacks the structured values when an error is fired and passes them through the structured values in the Error() call down the stack.

```
serrors.NewLogger(zapr.NewLogger(DefaultZapConfig(ctx, component).Build()))
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.23.0
 
 require (
 	github.com/Pallinder/go-randomdata v1.2.0
+	github.com/go-logr/logr v1.4.2
 	github.com/go-logr/zapr v1.3.0
 	github.com/imdario/mergo v0.3.16
 	github.com/onsi/ginkgo/v2 v2.23.4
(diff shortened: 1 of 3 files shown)


rschalo (approved):
/lgtm

PR #146, opened 2025-04-14
chore: Modify error handler middleware to return structured errors

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

Update the middleware to handle parsing out all details from AWS API errors, not just the request ID

### DIFF
--- a/aws/middleware/middleware.go
+++ b/aws/middleware/middleware.go
@@ -2,23 +2,41 @@ package middleware
 
 import (
 	"context"
+	"errors"
 
 	"github.com/aws/aws-sdk-go-v2/aws/transport/http"
+	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/middleware"
 	"github.com/awslabs/operatorpkg/serrors"
-	"github.com/samber/lo"
 )
 
-const AWSRequestIDLogKey = "aws-request-id"
+const (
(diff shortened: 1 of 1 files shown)


aws/middleware/middleware.go
```diff
@@ -2,23 +2,39 @@ package middleware
...
+			if v, ok := temp.(*http.ResponseError); ok {
+				values = append(values, AWSRequestIDLogKey, v.RequestID, AWSStatusCodeLogKey, v.Response.StatusCode)
+			}
+			if v, ok := temp.(*smithy.GenericAPIError); ok {
+				values = append(values, AWSErrorCodeLogKey, v.Code)
 			}
+			temp = errors.Unwrap(temp)
 		}
-		return out, metadata, err
+		return out, metadata, lo.Ternary(err != nil, serrors.Wrap(err, values...), nil)
```
jmdeal, line 38: nit: do we want to just short-circuit at the top if error == nil and then append values into the error directly?

jmdeal (approved):
One style nit, but otherwise LGTM

aws/middleware/middleware.go
jonathan-innis: Fair point

jmdeal (approved):
LGTM 🚀

PR #147, opened 2025-04-15
chore: Support structured errors for multierr

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

Support structured errors for `multierr.Append` and `multierr.Combine`. `multierr` is more challenging because multiple structured errors can be packed together. Additionally, it's lucky that errors in a multierr have the same values. As a result, `serrors.UnwrapValues()` combines elements with the same key together into a single array of values mapped to a single key. We keep the keyed values inside the call to `Error()` to ensure that the context around the key/value is retained

### DIFF
--- a/serrors/serrors.go
+++ b/serrors/serrors.go
@@ -2,12 +2,18 @@ package serrors
 
 import (
 	"errors"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/samber/lo"
+	"go.uber.org/multierr"
 )
 
 // Error is a structured error that stores structured errors and values alongside the error
 type Error struct {
 	error
-	keysAndValues []any
+	keysAndValues map[string]any
 }
 
 // Unwrap returns the unwrapped error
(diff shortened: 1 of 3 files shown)


serrors/serrors.go
```diff
@@ -17,38 +23,68 @@ func (e *Error) Unwrap() error {
...
+		v := e.keysAndValues[k]
+		elems = append(elems, fmt.Sprintf("%s=%v", k, v))
+	}
+	return fmt.Sprintf("%s (%s)", e.error.Error(), strings.Join(elems, ", "))
 }
 
 // WithValues injects additional structured keys and values into the error
 func (e *Error) WithValues(keysAndValues ...any) *Error {
-	e.keysAndValues = append(e.keysAndValues, keysAndValues...)
+	for i := 0; i < len(keysAndValues); i += 2 {
```
jmdeal, line 38: It's not specific to this PR, but while we're here should we add a panic if an odd number of elements are passed, same as in `Wrap`?

serrors/serrors.go
```diff
@@ -17,38 +23,68 @@ func (e *Error) Unwrap() error {
...
 	}
-	return values
+	keys := lo.Keys(values)
+	sort.Strings(keys) // sort keys so we always get a consistent ordering
+	for _, k := range keys {
+		v := values[k]
+		if len(v) == 1 {
+			res = append(res, k, v[0])
+		} else {
+			res = append(res, fmt.Sprintf("%ss", k), v)
```
jmdeal, line 86: To match the previous behavior, should we have duplicate pairs of keys and values rather than a single key followed by multiple values? Let's say we have a key (`foo`) with multiple values from different errors (`bar` and `baz`). Right now the result would be `[]any{"foos", []any{"bar", "baz"}}`, but do we want it to actually be `[]any{"foo", "bar", "foo", "baz"}`? I think what you have currently makes sense given the context I know it's used in, but wanted to sanity check.

jonathan-innis: I'd rather than the top one personally, that's going to lead to a much more readable structured log

serrors/serrors.go
jonathan-innis: Good callout -- added the validation

serrors/serrors.go
```diff
@@ -17,38 +23,76 @@ func (e *Error) Unwrap() error {
...
+	lo.Must0(validateKeysAndValues(keysAndValues))
+	for i := 0; i < len(keysAndValues); i += 2 {
+		e.keysAndValues[keysAndValues[i].(string)] = keysAndValues[i+1]
+	}
 	return e
 }
 
 // Wrap wraps and existing error with additional structured keys and values
 func Wrap(err error, keysAndValues ...any) error {
+	lo.Must0(validateKeysAndValues(keysAndValues))
```
jmdeal, line 47: nit: do we want to just construct a structured error with on values and call WithValues? Removes a bit of duplication, this is simple enough though so not a big deal either way IMO.

jmdeal (approved):
LGTM 🚀

PR #155, opened 2025-05-11
perf: Use partial unstructured converter to reduce memory consumption

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

This change ensures we only convert partial portions of the objects when we convert them to Unstructured so we don't use large portions of memory for really large objects.

### DIFF
--- a/unstructured/unstructured.go
+++ b/unstructured/unstructured.go
@@ -13,6 +13,13 @@ import (
 // This is more memory efficient than using runtime.DefaultUnstructuredConverter since that requires the full
 // object to be converted and stored before extracting specific values from that object
 func ToPartialUnstructured(obj interface{}, fieldPaths ...string) map[string]interface{} {
+	if u, ok := obj.(unstructured.Unstructured); ok {
+		obj = u.UnstructuredContent()
+	}
+	if u, ok := obj.(*unstructured.Unstructured); ok {
+		obj = u.UnstructuredContent()
+	}
+
 	result := make(map[string]interface{})
 	for _, fieldPath := range fieldPaths {
 		_ = extractNestedField(obj, result, lo.Filter(strings.Split(fieldPath, "."), func(s string, _ int) bool { return s != "" })...)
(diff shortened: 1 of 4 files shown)


unstructured/unstructured.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	// Intermediate map — recurse
+	childMap := make(map[string]interface{})
+	err := extractNestedField(val.Interface(), childMap, field[1:]...)
+	if err != nil {
+		return err
+	}
+	// Merge into parent map
+	if existing, exists := result[field[0]]; exists {
+		if m, ok := existing.(map[string]interface{}); ok {
```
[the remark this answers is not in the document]
jonathan-innis: It's a good point -- given the field check, I'm with you that I'm just going to assume it's always a map to simplify

unstructured/unstructured.go
```diff
@@ -61,20 +62,17 @@ func extractNestedField(obj interface{}, result map[string]interface{}, field ..
...
-				m[k] = v
-			}
-		}
-	} else {
-		result[field[0]] = childMap
+	if _, ok := result[field[0]]; !ok {
+		result[field[0]] = map[string]interface{}{}
+	}
+	for k, v := range childMap {
+		result[field[0]].(map[string]interface{})[k] = v
```
jmdeal, line 75: Still wondering if we want an explicit check here to verify that the type is a map. We'll panic here if it's not already, but it would be good to have a more explicit message IMO.

unstructured/unstructured.go
```diff
@@ -13,6 +13,13 @@ import (
 // This is more memory efficient than using runtime.DefaultUnstructuredConverter since that requires the full
 // object to be converted and stored before extracting specific values from that object
 func ToPartialUnstructured(obj interface{}, fieldPaths ...string) map[string]interface{} {
+	if u, ok := obj.(unstructured.Unstructured); ok {
+		obj = u.UnstructuredContent()
+	}
+	if u, ok := obj.(*unstructured.Unstructured); ok {
+		obj = u.UnstructuredContent()
+	}
```
jmdeal, line 21: Can you explain to me why this is needed?

jmdeal (approved):
LGTM 🚀

PR #162, opened 2025-06-12
fix: delete deprecated metrics

jigisha620:
*Issue*
https://github.com/aws/karpenter-provider-aws/issues/8159

*Description of changes:*
We were not deleting the old metrics because DeletePartialMatch() call was being made for the old metrics with additionalLabels that do not exist on the old metrics. This change updates the DeletePartialMatch() call for old metrics to not have additional labels.


status/controller_test.go
```diff
@@ -660,6 +665,9 @@ var _ = Describe("Controller", func() {
 		Expect(GetMetric("operator_customobject_status_condition_current_status_seconds", lo.Assign(conditionLabels(test.ConditionTypeFoo, metav1.ConditionFalse), map[string]string{"operator_pkg_key1": "value1", "operator_pkg_key2": "value2", "operator_pkg_key3": "value3"}))).To(BeNil())
 		Expect(GetMetric("operator_customobject_status_condition_current_status_seconds", lo.Assign(conditionLabels(test.ConditionTypeFoo, metav1.ConditionUnknown), map[string]string{"operator_pkg_key1": "value1", "operator_pkg_key2": "value2", "operator_pkg_key3": ""})).GetGauge()).To(BeNil())
 		Expect(GetMetric("operator_customobject_status_condition_current_status_seconds", lo.Assign(conditionLabels(test.ConditionTypeFoo, metav1.ConditionUnknown), map[string]string{"operator_pkg_key1": "value1", "operator_pkg_key2": "value2", "operator_pkg_key3": "value3"})).GetGauge().GetValue()).ToNot(BeZero())
+
+		Expect(GetMetric("operator_status_condition_count", conditionLabels(test.ConditionTypeFoo, metav1.ConditionUnknown)).GetGauge().GetValue()).To(BeEquivalentTo(0))
```
jonathan-innis, line 669: Let's briefly sync on this -- I'm not clear based on this test why this metric would be deleted

jigisha620: It happens because the additionalGaugeLabels != observedGaugeLabels on this line - https://github.com/awslabs/operatorpkg/blob/486985a7610cc5326bbb2872eab44e4dd8a6b902/status/controller.go#L278

jigisha620: Discussed offline, deprecated metrics should not be deleted on additional label change.

status/controller.go
```diff
@@ -275,21 +275,23 @@ func (c *Controller[T]) reconcile(ctx context.Context, req reconcile.Request, o
 	}
 
 	for _, observedCondition := range observedConditions.List() {
-		if currentCondition := currentConditions.Get(observedCondition.Type); currentCondition == nil || currentCondition.Status != observedCondition.Status || currentCondition.Reason != observedCondition.Reason || !maps.Equal(c.toAdditionalGaugeMetricLabels(o), observedGaugeLabels) {
-			c.deletePartialMatchGaugeMetric(c.ConditionCount, ConditionCount, lo.Assign(map[string]string{
+		currentCondition := currentConditions.Get(observedCondition.Type)
+		conditionChanged := currentCondition == nil || currentCondition.Status != observedCondition.Status || currentCondition.Reason != observedCondition.Reason
+		if conditionChanged || !maps.Equal(c.toAdditionalGaugeMetricLabels(o), observedGaugeLabels) {
```
jonathan-innis, line 280: nit: So technically what you can do here is you can only apply the additionalGaugeMetricLabels to the delete when you actually need to include them in the delete -- basically, add them when they change, don't add them when they don't and I think everything should work as it should

status/controller.go
```diff
@@ -282,14 +282,14 @@ func (c *Controller[T]) reconcile(ctx context.Context, req reconcile.Request, o
 				pmetrics.LabelType:         observedCondition.Type,
 				MetricLabelConditionStatus: string(observedCondition.Status),
 				pmetrics.LabelReason:       observedCondition.Reason,
-			}, observedGaugeLabels))
```
jonathan-innis, line 285: nit: Consider adding a comment here so it's clear why we are doing this

jonathan-innis (approved):
LGTM 🚀

status/controller.go
```diff
@@ -276,20 +276,25 @@ func (c *Controller[T]) reconcile(ctx context.Context, req reconcile.Request, o
 
 	for _, observedCondition := range observedConditions.List() {
 		if currentCondition := currentConditions.Get(observedCondition.Type); currentCondition == nil || currentCondition.Status != observedCondition.Status || currentCondition.Reason != observedCondition.Reason || !maps.Equal(c.toAdditionalGaugeMetricLabels(o), observedGaugeLabels) {
+			// We want to check if the additional lables on the object has changed, and if so, delete the metrics with the old lables.
```
jonathan-innis, line 279:
```suggestion
			// We want to check if the additional labels on the object has changed, and if so, delete the metrics with the old labels.
```

nit

PR #164, opened 2025-06-19
feat: make status histogram buckets configurable

aidan-canva:
*Issue #, if available:*

*Description of changes:*

Following breadcrumbs from Karpenter.  

We are drilling into attributing costs to the lifecycle of Nodes (EC2 instances) within our platform.  The Karpenter project uses `awslabs/operatorpkg` to emit metrics on status condition changes for CRDs like `NodeClaim` and `NodePool` - exactly what we are after.

Unfortunately, these metrics are emitted as histograms with non configurable buckets (defaults to `prometheus.DefBuckets`).  As objects like `NodeClaim` can take seconds to many minutes (or longer) to transition state, the default max bucket of 5seconds makes these metrics less useful.

This PR allows consumers to optionally configure the histogram metrics buckets, keeping the existing default behavior.


jonathan-innis (approved):
Nice! Thanks for the change! 🎉

PR #174, opened 2025-08-08
Adding degraded status condition

sarthakkothari:
*Description of changes:*

Today we only support ConditionTrue, ConditionFalse and ConditionUnknown. We also require a condition that marks Condition as degraded for a user error. This PR helps us mark status conditions as degraded.

Degraded also takes precedence over False and unknown when setting the root condition.

### DIFF
--- a/status/condition.go
+++ b/status/condition.go
@@ -26,6 +26,10 @@ const (
 	ConditionSucceeded = "Succeeded"
 )
 
+// ConditionDegraded is a custom extension to the standard Kubernetes condition statuses.
+// It represents a state where the resource is operational but in a degraded state.
+const ConditionDegraded metav1.ConditionStatus = "Degraded"
+
 // Condition aliases the upstream type and adds additional helper methods
 type Condition metav1.Condition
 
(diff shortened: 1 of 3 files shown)


status/condition.go
```diff
@@ -26,6 +26,10 @@ const (
 	ConditionSucceeded = "Succeeded"
 )
 
+// ConditionDegraded is a custom extension to the standard Kubernetes condition statuses.
+// It represents a state where the resource is operational but in a degraded state.
```
sarthakkothari, line 30:
```suggestion
// It represents a state where the resource is degraded for a known reason
```

status/condition.go
```diff
@@ -26,6 +26,10 @@ const (
...
+const ConditionDegraded metav1.ConditionStatus = "Degraded"
```
jonathan-innis, line 31: I'd be wary of supporting concepts that aren't known concepts in upstream K8s. 

Why is this not the same thing as "Unknown"? That's sort of the implication when you don't have the info in the same way as Degraded.

status/condition.go
jonathan-innis: I struggle to differentiate this from the "False" state -- this is also a scenario when the status condition is probably "not working as it should" i.e. "degraded" AND this is also a known state.

Also, I'll add that this status being a new addition to True/False/Unknown means that it needs to make sense in the context of basically every status condition that we could write -- yet again, I feel like I would really struggle to determine whether I should mark a status condition as False or Degraded based on the failure mode

jonathan-innis:
We had a discussion offline and decided that this was not needed. Basically, we discussed the fact that any server-side issue that doesn't require user intervention should leave status conditions in Unknown -- a True state means that things are healthy for the system and a False state implies that there is some known error that involves user misconfiguration and that involves some form of user intervention.

PR #180, opened 2025-08-26
fix: Update ExpectApplied to handle conflicts and to only update status when needed

jonathan-innis:
*Issue #, if available:*

*Description of changes:*

Update `ExpectApplied` so that it is resilient to conflicts as well as can avoid making calls to update the status when the status hasn't changed OR the status wasn't set (and therefore, doesn't want to be updated) by the caller


jmdeal (dismissed):
LGTM 🚀

PR #182, opened 2025-08-31
test: remove isStatusSet check

rlanhellas:
To fix test error on this PR https://github.com/kubernetes-sigs/karpenter/pull/2368

*Description of changes:*

When calling `ExpectApplied` it creates the resource and update the status according to k8s flow, i.e: For POD resource, it enters in `Pending` state. But instead we want to keep status sent by user in test and don't override it. It can be empty, it means the status should kept empty too, so removing `isStatusSet` will let status kept empty when set by developer in test.


rlanhellas:
closing this PR since we don't want to override status in operatorpkg, if we need to do it, should be done on upstream project not operatorpkg tests

PR #204, opened 2026-04-07
feat: support injected clock in ConditionSet for testability

jamesmt-aws:
## Summary

`ConditionSet.Set()` uses `metav1.Now()` for `LastTransitionTime`, which returns real wall clock time. Controllers that check elapsed time since a condition was set using an injected fake clock get incorrect results because the condition timestamp and the controller clock diverge.

Add a `WithClock` option to `ConditionTypes.For()`:

```go
// Production (default, no change required)
conditions := status.NewReadyConditions(...).For(object)

// Tests with fake clock
conditions := status.NewReadyConditions(...).For(object, status.WithClock(fakeClock))
```

Defaults to `clock.RealClock{}`, so existing callers are unaffected.

## Motivation

This fixes a class of test flakes in downstream projects. In kubernetes-sigs/karpenter, the termination controller checks `c.clock.Since(condition.LastTransitionTime)` using a fake clock, but `LastTransitionTime` was stamped with `metav1.Now()`. The test "should not finish draining until minDrainTime has passed" fails deterministically because the fake clock and real clock diverge (kubernetes-sigs/karpenter#2951).

The downstream workaround is to track timestamps separately using the injected clock, duplicating what `LastTransitionTime` already provides. This PR fixes the root cause.

## Changes

- `condition_set.go`: Add `clock` field to `ConditionSet`, `ForOptions`/`ForOption`/`WithClock` types, `now()` helper. Replace two `metav1.Now()` calls with `c.now()`.
- `condition_set_test.go`: Test that `WithClock` controls `LastTransitionTime` on set, transition, and same-status update.

## Test plan

- [x] All existing tests pass (no behavioral change without `WithClock`)
- [x] New test verifies fake clock controls `LastTransitionTime`

### DIFF
--- a/status/condition_set.go
+++ b/status/condition_set.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/samber/lo"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/clock"
 )
 
 // ConditionTypes is an abstract collection of the possible ConditionType values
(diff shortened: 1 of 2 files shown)


tzneal (approved):
lgtm

PR #205, opened 2026-04-09
chore: migrate status controller to events.EventRecorder API

jamesmt-aws:
## Summary

- Migrate `status.NewController` and `status.NewGenericObjectController` to take `k8s.io/client-go/tools/events.EventRecorder` instead of the deprecated `k8s.io/client-go/tools/record.EventRecorder`.
- Translate the three internal `eventRecorder.Event(...)` call sites to `Eventf(...)` with action strings `Finalize` and `TransitionCondition`.
- Tests use `events.FakeRecorder`, which emits the same `<type> <reason> <note>` channel format as the old fake — no assertion changes needed.

## Motivation

controller-runtime v0.23 deprecated `manager.GetEventRecorderFor` and the legacy `record.EventRecorder` API in favor of the newer `events.EventRecorder` (`k8s.io/client-go/tools/events`). The deprecation notice says the old API will be removed in a future release.

Downstream consumers of operatorpkg cannot migrate off `GetEventRecorderFor` as long as `status.NewController` requires the legacy `record.EventRecorder`. In kubernetes-sigs/karpenter#2951 we currently have to add `//nolint:staticcheck` directives at every call site to suppress the SA1019 warning, which papers over the deprecation rather than addressing it.

This PR unblocks karpenter (and any other downstream consumer) to fully migrate to the new events API.

## Breaking change

This changes the public signature of:
- `status.NewController[T Object](client, eventRecorder events.EventRecorder, ...)`
- `status.NewGenericObjectController[T client.Object](client, eventRecorder events.EventRecorder, ...)`

Callers will get a compile-time error with a clear migration path: replace `mgr.GetEventRecorderFor(name)` with `mgr.GetEventRecorder(name)`. There is no silent behavior change.

I'm open to the parallel-constructor approach (`NewControllerWithEventsRecorder` alongside the existing one) if you'd prefer a softer migration — happy to rework if so.

## Test plan

- [x] `go test ./...` passes locally
- [x] No test assertion changes required (FakeRecorder format is identical)
- [ ] CI green

### DIFF
--- a/status/controller.go
+++ b/status/controller.go
@@ -20,7 +20,7 @@ import (
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/tools/events"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


jamesmt-aws:
Closing in favor of #206 for design discussion. The downstream urgency dropped — kubernetes-sigs/karpenter#2951 turned out to be cleanly shippable with the two existing nolints, and the deprecation isn't a removal yet. The branch (https://github.com/jamesmt-aws/operatorpkg/tree/chore/migrate-events-recorder) is parked and ready to reopen as a PR once the approach is settled in #206.

PR #211, opened 2026-06-09
perf: fix GenericObjectController bypassing controller-runtime cache on every reconcile

GnatorX:
*Issue [3069](https://github.com/kubernetes-sigs/karpenter/pull/3069):*

GenericObjectController.Reconcile was calling kubeClient.Get with a *UnstructuredAdapter[T] object. Because UnstructuredAdapter embeds unstructured.Unstructured, it implements the runtime.Unstructured interface. controller-runtime's cached client checks for this interface in shouldBypassCache() and, when cacheUnstructured is not explicitly enabled, skips the cache and sends a direct HTTP GET to the API server.

Since GenericObjectController requeues every 10 seconds, this produced tens of thousands of uncached node (or other resource) GET requests per minute under normal idle operation — visible in audit logs and a meaningful load on the API server.

This matches the issue reported and fixed in karpenter upstream: https://github.com/kubernetes-sigs/karpenter/pull/3069

*Description of changes:*

Refactored Controller.reconcile into two internal helpers:

- cleanupNotFound — handles the not-found cleanup path (metric deletion, finalizer events)
- reconcileFound — handles all post-fetch processing (conditions, metrics, transitions)

GenericObjectController.Reconcile now:
1. Calls kubeClient.Get with a typed T object, which goes through the cache normally
2. Converts the result to proper unstructured form via runtime.DefaultUnstructuredConverter.ToUnstructured before wrapping in UnstructuredAdapter

You can see the GET requests to api server drop off a cliff after the change is deployed
Screenshot 2026-06-10 at 10 43 27 AM

### DIFF
--- a/status/controller.go
+++ b/status/controller.go
@@ -181,7 +181,28 @@ func (c *GenericObjectController[T]) Register(_ context.Context, m manager.Manag
 }
 
 func (c *GenericObjectController[T]) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	return c.reconcile(ctx, req, NewUnstructuredAdapter[T](object.New[T]()))
+	// Use a typed Get so controller-runtime uses its cache instead of bypassing it.
+	// UnstructuredAdapter implements runtime.Unstructured (via embedded unstructured.Unstructured),
+	// which causes controller-runtime to skip the cache and hit the API server directly when used
+	// with kubeClient.Get.
+	typedObj := object.New[T]()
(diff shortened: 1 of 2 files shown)


jmdeal (approved):
I've updated the title to call out that this is technically a breaking change since we'll now drop fields that aren't included in the go object definition that we wouldn't have previously. This wouldn't impact Karpenter, but could hypothetically impact some other consumer. Looks good to me though, thanks for looking into this!

PR #212, opened 2026-07-08
fix: don't fabricate a Ready condition for generic objects in metrics

DerekFrank:
**Problem**

`GenericObjectController` wraps objects in an `UnstructuredAdapter` whose `StatusConditions()` calls `NewReadyConditions(...).For(u)`. `For()` initializes any missing root/dependent condition type to `Unknown`. For an object that does **not** define a `Ready` condition, this fabricates a `Ready=Unknown` in the adapter's in-memory condition list, which the reconcile gauge loop then emits as a permanent metric:

```
operator_status_condition_count{type="Ready", status="Unknown", reason="AwaitingReconciliation"} 1
```

Nothing is ever persisted to the object (the adapter is a throwaway unstructured copy), and the synthetic `Ready` never rolls up, so it's stuck at `Unknown` forever — misleading noise for any object whose API doesn't use a `Ready` root condition.

**Fix**

Add a `WithObservedOnly()` `ForOption` that builds the `ConditionSet` from the conditions already present on the object, skipping initialization of absent root/dependent conditions to `Unknown`. The `UnstructuredAdapter` — a read-only reflection used only for metrics on objects the controller does not own — uses it, so it reports metrics for the conditions that actually exist and no longer fabricates a `Ready`.

Objects that own their status and drive reconcile (via `NewController[T]` with a typed `status.Object`) are unaffected: `For()` still initializes their declared conditions to `Unknown` as before.

**Testing**

Added a generic-controller test asserting no `Ready` metric is emitted for an object that defines only non-`Ready` conditions. Verified it fails before the fix (phantom `Ready=Unknown`) and passes after. Full `status` suite passes.

**Motivation**

Karpenter is adding `operator_status_condition_*` metrics for the `CapacityBuffer` CRD (whose conditions are `ReadyForProvisioning`/`Provisioning`/`LimitedByQuotas`, with no `Ready`) via `GenericObjectController`, and was seeing a permanent phantom `Ready=Unknown` series.

### DIFF
--- a/status/condition_set.go
+++ b/status/condition_set.go
@@ -54,6 +54,9 @@ type ConditionSet struct {
 // ForOptions configures a ConditionSet.
 type ForOptions struct {
 	Clock clock.Clock
+	// ObservedOnly builds the set from conditions already present, without
+	// initializing absent root/dependent conditions to Unknown. See WithObservedOnly.
+	ObservedOnly bool
 }
 
 // ForOption is a functional option for For.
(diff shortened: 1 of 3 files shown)


jmdeal (approved):
LGTM 🚀

PR #213, opened 2026-07-09
test: assert Ready metric still emits for generic objects that define one

DerekFrank:
Follow-up to #212.

#212 stopped the `GenericObjectController` from fabricating a phantom `Ready=Unknown` metric for objects that don't define a `Ready` condition, and added a test for that absent-`Ready` case.

This adds the complementary case: an object that carries a **real, persisted `Ready`** condition (e.g. `corev1.Node`, whose `Ready` is set by kubelet) must **still** emit its `Ready` metric. This confirms the `WithObservedOnly` change filters only *fabricated* conditions, not ones actually present on the object — i.e. no regression to the Node/NodeClaim/NodePool Ready metrics that legitimately flow through this path.

Test-only; passes against current main.


jmdeal (approved):
LGTM 🚀

PR #214, opened 2026-08-05
feat: add instrumented go-cache wrapper with Prometheus metrics

DerekFrank:
Add operatorpkg/cache, a near drop-in wrapper around patrickmn/go-cache that emits Prometheus metrics under the operator_cache_* subsystem:

- gets_total{name,result=hit|miss}
- adds_total{name,result=added|exists}  (exists = suppressed duplicate)
- evictions_total{name}                 (TTL expiry)
- deletes_total{name,type=explicit|flush}
- flushes_total{name}
- flush_size{name}                       (histogram, entries per flush)
- entries{name}                          (gauge)

The wrapper embeds *cache.Cache so every underlying method still works; only Get/Set/SetDefault/Add/Delete/Flush are overridden. Counter/gauge series are pre-initialized to 0 to avoid sparse series. The entries gauge is updated on mutation (never at scrape time) so scrapes do no cache work. A caller-supplied OnEvicted callback is chained, not replaced.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -1,0 +1,162 @@
+// Package cache wraps github.com/patrickmn/go-cache with Prometheus metrics.
+//
+// The wrapped Cache embeds *cache.Cache, so it is a near drop-in replacement:
+// every method of the underlying cache is still available, and only the
+// operations that carry an observable signal (Get, Set, SetDefault, Add,
+// Delete, Flush) are overridden to record metrics. Construct one with New,
+// passing a stable, low-cardinality name used as the "name" metric label.
+package cache
+
+import (
+	"time"
+
+	"github.com/patrickmn/go-cache"
+)
+
+// Item re-exports cache.Item so callers can migrate to this package without
+// also importing go-cache directly (e.g. for the map returned by Items()).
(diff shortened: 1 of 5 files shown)


cache/cache.go
```diff
@@ -0,0 +1,162 @@
...
+	for i := 0; i+1 < len(kv); i += 2 {
+		l[kv[i]] = kv[i+1]
+	}
+	return l
+}
+
+// Get records a hit or miss and delegates to the underlying cache.
+func (c *Cache) Get(k string) (interface{}, bool) {
+	v, ok := c.Cache.Get(k)
+	if ok {
```
ryan-mist, line 85: Theres a pretty small perf hit here because theres a bunch of allocations + the metric itself - it tripled the CPU usage on every `cache.Get()`

I don't think its that large, but since there is limited value in the metric it might not be worth it

> Test Scenario: 9 NodeClasses, 9 NodePools, 0 pods

Image

DerekFrank: Yea I think we can drop it

PR #216, opened 2026-08-12
feat(status): label transition_seconds with the destination status

DerekFrank:
*Issue #, if available:*

*Description of changes:*

The transition_seconds histogram only labeled the status a condition was LEFT (`status`), never the status it moved TO. A dwell in e.g. Ready=Unknown or Ready=False was therefore indistinguishable between "recovered to True" and any other outcome, so recovery-duration for a condition could not be isolated.

Add a `to_status` label to `operator[_<kind>]_status_condition_transition_seconds`. `status` still means the state being left (backward compatible); `to_status` is the state entered. Recovery duration is then transition_seconds filtered to status=False/Unknown, to_status=True. The recovery *rate* can be derived by comparing that against transitions_total entries into the unhealthy state.


ryan-mist (approved):
/lgtm

PR #218, opened 2026-08-28
feat(metrics): make metrics label-aware (carry their []Label)

DerekFrank:
The Prometheus* metric constructors now take []Label instead of []string, and each metric retains the Labels it was declared with (exposed via a Labels() method on the CounterMetric/GaugeMetric/ObservationMetric interfaces). The underlying Prometheus vector is still registered with the label Names.

This makes a metric the source of truth for its own dimensions — enabling downstream metrics-docs generation to read a metric's exact Labels (help + values) instead of resolving label names globally, and enabling runtime introspection/validation of the emitted label set.

BREAKING: NewPrometheus{Counter,Gauge,Histogram,Summary} signatures change from labelNames []string to labels []Label; callers pass []metrics.Label. The status/events metrics and the client-go metrics are updated accordingly.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/events/metrics.go
+++ b/events/metrics.go
@@ -15,9 +15,9 @@ func eventTotalMetric(objectName string) pmetrics.CounterMetric {
 			Name:      "event_total",
 			Help:      "The total of events of a given type for an object.",
 		},
-		[]string{
-			pmetrics.LabelType,
-			pmetrics.LabelReason,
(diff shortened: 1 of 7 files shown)


ryan-mist (approved):
/lgtm

PR #219, opened 2026-09-01
feat(metrics): add a required Stage field to metric constructors

DerekFrank:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/metrics/stage.go
+++ b/metrics/stage.go
@@ -1,0 +1,18 @@
+package metrics
+
+// Stage describes the API stability of a metric, mirroring the Kubernetes
+// metric stability levels. Declaring it lets a metrics documentation generator
+// surface which metrics are safe to depend on and which may still change.
+type Stage string
+
+const (
+	// Alpha marks an experimental metric that may change or be removed without
+	// notice.
+	Alpha Stage = "alpha"
+	// Beta marks a metric that is fairly stable but whose name, labels, or
+	// semantics may still change before it is promoted.
+	Beta Stage = "beta"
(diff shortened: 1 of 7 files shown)


metrics/stage.go
```diff
@@ -0,0 +1,18 @@
...
+
+const (
+	// Alpha marks an experimental metric that may change or be removed without
+	// notice.
+	Alpha Stage = "alpha"
+	// Beta marks a metric that is fairly stable but whose name, labels, or
+	// semantics may still change before it is promoted.
+	Beta Stage = "beta"
+	// GA marks a stable metric that is safe to depend on; breaking changes
+	// follow the usual deprecation process.
```
ryan-mist: nit: I think we should be explicit that in GA we are still free to add new dimensions

DerekFrank: I agree, maybe alpha means any changes are valid, beta means breaking changes to dimensions, GA means only additive changes?

ryan-mist: That sounds good to me

DerekFrank: Done in 1cbe26b — spelled out the per-stage contract in the const docs: **alpha** = any aspect (name/dimensions/values/the metric itself) may change or be removed without notice; **beta** = additive changes expected and breaking dimension changes (rename/remove) still permitted before promotion; **GA** = only additive — new dimensions may still be added, but existing ones aren't renamed/removed except via the usual deprecation process.

ryan-mist (approved):
/lgtm

